PR #2, opened 2020-10-30
Migrate to GraphQL and MongoDB

orkosinha:
This PR adds some basic functionality for migrating the backend to GraphQL and MongoDB.

### Overview
Creates GraphQL based backend for storing and accessing Cornell publications through a MongoDB database.
Articles are queryable by an article ID, by publisher, by a timestamp (certain number of articles published after a timestamp).
Publications are queryable by the publication slug stored in `publications.json` or can be fetched all at once.
Mutations are also available for liking articles (will be changed later) and refreshed (uses `db/rss-parser.ts`).

### Specifics
* Adds resolvers for Articles and Publications `resolvers/*`
* Adds Publication class that is used for GraphQL schema definition `entities/Publication.ts`
* Adds Article class that works as MongoDB model and GraphQL schema definition `entities/Article.ts`
* Adds function for parsing publication feeds and creating Article models `db/rss-parser.ts`
* Adds connection to MongoDB database `db/DBConnection.ts`
* Migrates linting to using `eslint-config-airbnb-typescript-prettier`

Tested using GraphQL playground and MongoDB Compass.

Thanks @manishsaha for figuring out how to fix the linting issue with React.

-- Sorry this PR is extremely fat, a lot of the code is boilerplating for GraphQL and MongoDB apologies


src/db/rss-parser.ts
```diff
@@ -0,0 +1,45 @@
...
+    nameToSlugMap[publication['rss-name']] = publication.slug;
+    return publication.feed;
+  });
+
+  const feedRequests = feedSources.map((feed) => {
+    return parser.parseURL(feed);
+  });
+
+  const articlePromises = Promise.all(feedRequests).then((feeds) => {
+    const articles = new Array<Article>();
```
grebneerg, line 21: Mostly stylistic suggestion: if you are using node >= 11 (current lts is 14), you could use [Array.prototype.flat()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat) on the result of your nested maps instead of continuously mutating the `articles` variable (the `const` declaration of which is misleading if you ask me).

grebneerg (approved):
You might want to think about adding some tests for this, but otherwise it looks good!

src/entities/Article.ts
```diff
@@ -0,0 +1,34 @@
...
+  @Field(() => ID)
+  id: string;
+
+  @Field()
+  @Property()
+  title: string;
+
+  @Field()
+  @Property()
+  publication: string;
```
cswenberg, line 15: i've never done a TS backend project so maybe I just don't know the syntax, but shouldn't this be a FK to reference the other Publication entity?

src/entities/Publication.ts
```diff
@@ -0,0 +1,22 @@
...
+  slug: string;
+
+  @Field()
+  bio: string;
+
+  @Field()
+  feed: string;
+
+  @Field({ nullable: true })
+  img: string;
```
cswenberg, line 15: inconsistent naming with `imageURL` from `Article.ts`

src/entities/Publication.ts
```diff
@@ -0,0 +1,22 @@
...
+@ObjectType({ description: 'The Publications Model' })
+export default class Publication {
+  @Field()
+  slug: string;
+
+  @Field()
+  bio: string;
+
+  @Field()
+  feed: string;
```
cswenberg, line 12: same thing about the Article <-> Publication reference, why is this a string?

src/entities/Publication.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  @Field({ nullable: true })
+  img: string;
+
+  @Field()
+  name: string;
+
+  @Field()
+  url: string;
```
cswenberg, line 21: if I were to be extra picky, I'd say this should be `websiteURL` to match the style from `Article.ts`

src/resolvers/PublicationResolver.ts
```diff
@@ -0,0 +1,31 @@
...
+  @Query((_returns) => [Publication], { nullable: false })
+  async getAllPublications() {
+    const publicationArray = publicationsDB.map((publication) =>
+      Object.assign(new Publication(), {
+        slug: publication.slug,
+        bio: publication.bio,
+        feed: publication.feed,
+        img: publication.img,
+        url: publication.url,
+        name: publication.name,
```
cswenberg, line 18: abc?

utils/createDummyData.ts
```diff
@@ -0,0 +1,71 @@
...
+        title, publication, articleURL, imageURL, date, likes,
+      } = article;
+      const newArticle = await ArticleModel.create({
+        title,
+        publication,
+        date,
+        imageURL,
+        articleURL,
+        likes,
+      });
```
cswenberg, line 62: i'm guessing this can't be abc'd because the shorthand syntax expects the arguments to come in the same order as the fields are defined in the entity file?

cswenberg (commented):
I don't want to block this PR to get the caboose moving but could we also get a more populated README (things like setup and such to begin)? Not essential right now, but we should try to build the habit of new documentation for every contribution we make. 

Also non-essential but consider rebasing to clean up your commit history and fix some of the names (some typo and convention things).

src/entities/Publication.ts
cswenberg: also if this is supposed to represent the articles belonging to this publication, wouldn't the name `articles` be more direct and clear? to me, a feed is a more generalistic collection of articles that are pulled together from multiple sources, so in that sense, it's awkward to say this "feed" belongs to this publication vs. these "articles" belong to this publication.

aliciaxw: agreed

utils/createDummyData.ts
```diff
@@ -0,0 +1,71 @@
...
+    {
+      title: 'orko and tedi backend ',
+      publication: 'dailysun',
+      artilceURL: 'www.orko.com',
+      imageURL: 'www.orkopics.com',
+      date: new Date('10-14-2020'),
+      likes: 0,
+    },
+    {
+      title: 'appdev router sucks ---> let me tell you why',
```
aliciaxw, line 17: lol

src/entities/Article.ts
```diff
@@ -0,0 +1,34 @@
+import { Field, ID, ObjectType } from 'type-graphql';
+import { prop as Property, getModelForClass } from '@typegoose/typegoose';
+
+@ObjectType({ description: 'The Article Model' })
+export class Article {
+  @Field(() => ID)
+  id: string;
```
aliciaxw, line 7: there's been a push to name the id field `uuid` and not `id`

aliciaxw (changes requested):
i would address conner's comment about Article <-> Publication reference in your entities and the minor comments about consistent naming (img vs imageURL), but things like tests and readme can be done in separate prs! in general, amazing work, and really excited to see appdev's first nosql database :-) 🚀

src/entities/Article.ts
orkosinha: Not with the current design because Publication isn’t a table being stored in the database. However, I can change this to being a MongoDB entity and adding a new table for publishers. The motivation behind the design was that publisher data is static, so there wasn’t too much of a need to make a table for it. This isn’t really scalable though for more features in the future so it may be worth changing.

src/entities/Publication.ts
orkosinha: Fixed

src/entities/Article.ts
aliciaxw: @orkosinha let's keep what you have so far and make a github issue for it to change it after this PR is merged tbh, as long as we're aware there's a more scalable solution

src/db/rss-parser.ts
orkosinha: fixed

src/entities/Publication.ts
orkosinha: fixed, mostly was an issue with `rss-parser.ts` but I think it reads better now

src/entities/Publication.ts
orkosinha: Agreed, fixed

src/resolvers/PublicationResolver.ts
orkosinha: Fixed

utils/createDummyData.ts
orkosinha: ArticleModel is never using create like this in our code, so it isn't an issue.

src/entities/Article.ts
orkosinha: Fixed

aliciaxw (approved):
great edits, LGTM!!! go volume!!!!!!

PR #3, opened 2020-11-08
Add image parsing from RSS

orkosinha:
Adds image parsing through [cheerio](https://cheerio.js.org/). 

Article data in HTML is contained in RSS through either the `content` or `content:encoded` tags with `content:encoded` being guaranteed (at least in the case of these RSS feeds) to have HTML content. `content:encoded` may not be specified in an RSS feed, but `content` is guaranteed (by specification) and this may contain HTML content. So we check `content:encoded` first to see if it exists and otherwise fall back to `content` for an HTML string. This is then searched with cheerio to find the first image tag (when searching through an RSS feed content, image data is always supposed to be relevant to the article so we avoid mistakes with selecting extraneous content). 

Not all articles will have images, so we fall back to empty string in the cases where no image is associated with the article. This could be because either the publication isn't really about images in articles (The Undergraduate Law & Society Review never has any) or there is no image associated with the article by design (op-ed pieces). 

A possible way to improve this parsing method is to make scrappers that are publication specific. Cornell Daily Sun uses an alternative method for their main images in articles, and all this (usually) gets are images within the article and not the main header image. However, this strategy works for smaller publications as their header images are contained within the RSS feed.

Also created a utility to merge all RSS feeds into one feed in the `utils` folder and a PR template. Also I messed up last time and accidentally merged the last PR. I disabled this in the repo settings and now we should only be able to squash and merge.

## Testing

This method provides the highest percentage of articles with images (~68% of articles from today). I'm not 100% sure all the images are relevant to the article, but all of the ones I've checked are. I really hope we don't put up any bad images.

### DIFF
--- a/src/db/rss-parser.ts
+++ b/src/db/rss-parser.ts
@@ -1,4 +1,5 @@
 import Parser from 'rss-parser';
+import cheerio from 'cheerio';
 import { Article } from '../entities/Article';
 import publicationsJSON from '../../publications.json';
 
@@ -16,6 +17,12 @@ const getRecentArticles = async (): Promise<Article[]> => {
     return parser.parseURL(articles);
   });
 
+  const parseImage = function (content) {
+    const $ = cheerio.load(content);
+    const img = $('img').attr('src');
+    return img || '';
+  };
+
   const articlePromises = Promise.all(parsedPublications).then((publications) =>
     publications
       .map((publication) =>
@@ -23,7 +30,9 @@ const getRecentArticles = async (): Promise<Article[]> => {
           Object.assign(new Article(), {
             articleURL: article.link,
             date: new Date(article.pubDate),
-            imageURL: '',
+            imageURL: parseImage(
+              article['content:encoded'] ? article['content:encoded'] : article.content,
+            ),
             likes: 0,
             publication: nameToSlugMap[publication.title]
               ? nameToSlugMap[publication.title]
--- a/utils/rss-feed.ts
+++ b/utils/rss-feed.ts
@@ -1,25 +1,19 @@
 import Parser from 'rss-parser';
 import dotenv from 'dotenv';
 import express from 'express';
+import publicationsJSON from '../publications.json';
 
-const PORT = process.env.APP_PORT || 5000;
+const PORT = 5000;
 
 // load the environment variables from the .env file
 dotenv.config({
     path: '.env',
 });
 
-const FEED_LIST = ['https://cornellsun.com/feed/',
-                   'http://cunooz.com/?feed=rss2',
-                   'http://theadvocatecornell.com/feed/',
-                   'https://medium.com/feed/guac-magazine',
-                   'https://www.cornellclaritas.com/blog?format=rss',
-                   'https://www.culsr.org/?format=rss',
-                   'https://www.cremedecornell.net/blogposts?format=rss',
-                   'https://www.slopemedia.org/all?format=rss',
-                   'https://www.thecornellreview.org/feed/',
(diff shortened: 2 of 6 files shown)


src/db/rss-parser.ts
```diff
@@ -16,14 +17,22 @@ const getRecentArticles = async (): Promise<Article[]> => {
     return parser.parseURL(articles);
   });
 
+  const parseImage = function (content) {
+    const $ = cheerio.load(content);
+    const img = $('img').attr('src');
```
cswenberg, line 22: I've never seen `$` used for naming const/vars/funcs before, is this common practice for using cheerio or typescript? Just seems unfamiliar and not that expressive?

utils/rss-feed.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    const feedRequests = FEED_LIST.map((feed) => {
+        return parser.parseURL(feed);
+    });
+
+    Promise.all(feedRequests).then((response) => {
+        res.setHeader('Access-Control-Allow-Origin', '*');
+        res.header('Access-Control-Allow-Methods', 'GET');
+        res.json(response);
+    });
```
cswenberg, line 31: Not too familiar with `async-await` syntax vs. Promises but I do know the former is a newer feature. Is there a reason we're choosing to use `.then()` syntax instead of `await`? Also what happens if one of the feed requests goes wrong (we don't have any `.catch()` defined)?

cswenberg (commented):
Your PR description was super helpful, consider turning it into part of the documentation given you've already written it! RSS parsing is a new thing for the team so it would be a huge help for people to easily understand the nuances of `content:encoded` and all the image/no-image options you've described and thought through. Don't let that good insight only stay inside your noggin' 😉

src/db/rss-parser.ts
orkosinha: I think this is how cheerio recommends using it just because it mocks jQuery with line 22 being the same between both cheerio and jQuery. This is syntax is also used with pear-backend and getting the majors in Web-Scraper.ts

utils/rss-feed.ts
orkosinha: I didn’t spend too much time with this code, this was just a starter file rss-parser had on their docs site and modified to see all of the articles in one feed. I could change it here if it’s considered better practice (and a few other places .then is used)

aliciaxw: +1 to this!

aliciaxw: `Promise.all` is fine here imo; in the case that any of the feed requests fail, we get a fast fail which we wouldn't with multiple awaits

utils/rss-feed.ts
```diff
@@ -0,0 +1,32 @@
...
+    return publication.feed;
+  });
+
+const app = express();
+
+app.get('/', (req, res) => {
+    const parser = new Parser();
+
+    const feedRequests = FEED_LIST.map((feed) => {
+        return parser.parseURL(feed);
```
aliciaxw, line 24: shouldn't we be awaiting here? unsure how `feedRequests` are promises

aliciaxw (commented):
have some questions about the rss-feed code, if we can address then it's good to go!

utils/rss-feed.ts
orkosinha: `parseURL` returns a `Promise<Parser.Output>` which is then put into the `feedRequests`.  Also, if we use `Promise.all()` doesn't the input have to be an array of promises? So if we make this await, wouldn't it resolve to `Parser.Output`?

tedim52 (approved):
LGTM.

PR #4, opened 2020-11-17
Add a README and trending articles query, Refactor Publication Model

tedim52:
## Overview

-Added a README file
-Added getTrendingArticles Query
-Refactored Publication Model

I initially was going to make these three separate atomic PR's but then decided to make it one because I thought each of them had only really small changes and now I realize I probably still should have made them three PR's.

## Changes Made
-Added a query to get trendiness articles based on a simple "trendiness function"
-Added a test case to test the trendiness comparator
-Made Publications into a schema in the database so that total shoutouts can be stored and manipulated
-Removed "slug" field and made it so all publications are now queried by it's ObjectId 
-Created Repos for Article and Publication model to keep minimal logic within resolvers, Orko and I made this decision to make testing easier down the line and the resolvers easier to read
-Made it so that all files -except for initially loading publication data in the database - pulls publication data from database and not static JSON file

## Test Coverage

Tested queries extensively through GraphQL playground.
Created a test case for trendiness comparator.

## Related PRs or Issues

A side effect of storing the number of shoutouts for Publications and Articles without tracking user accounts is that should someone find the query, they can manually increment the number of shoutouts. We need to find some way to fix this. We initially were going to do this by storing User info UUID, bookmarked articles, etc. but we opted not to because the frontend could easily store this info.


src/app.ts
```diff
@@ -22,6 +22,9 @@ const main = async () => {
 
   await dbConnection();
 
+  //Prefill publication data
```
chalo2000, line 25: space between // and Prefill

src/repos/ArticleRepo.ts
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * Computes and returns the trending articles in the database.
+ *
+ * Trendiness is computed by taking the number of total shoutouts an article
+ * has received and dividing it by the number of days since its been published.
+ *
+ * @function
+ * @param {number} limit - number of articles to retrieve.
+ * @param {Date} since - retrieve articles after this date.
```
chalo2000, line 42: it looks like the param type is string but you convert it to a Date

tedim52: Yup, you're right will make the change!

utils/createDummyData.ts
```diff
@@ -1,71 +1,68 @@
...
-      likes: 0,
+      shoutouts: 0,
     },
     {
       title: 'Cooking w Cornell Hotel School',
-      publication: 'creme',
-      articleURL: 'www.ne',
-      imageURL: 'www.pic',
+      publicationID: '5fac75af7f68e1f1a4ae8b1b',
+      articleURL: 'www.new',
```
chalo2000, line 26: www.new.com

tests/articlerepo.test.ts
```diff
@@ -0,0 +1,101 @@
+import MockDate from 'mockdate';
+import { compareTrendiness } from '../src/repos/ArticleRepo';
+
+describe("testing getTrendingArticles query", () => {
+  beforeAll(() => {
+    //Set present time to a fixed day to compute trendiness from
```
chalo2000, line 6: // Set

utils/createDummyData.ts
chalo2000, line 26: www.new.com

chalo2000 (approved):
noice first PR 🔥

src/repos/PublicationRepo.ts
```diff
@@ -0,0 +1,47 @@
...
+    const { bio, rssURL, imageURL, name, websiteURL, rssName } = publication;
+    publications.push(Object.assign(new Publication(), {
+      _id: new ObjectId().toString(),
+      bio,
+      rssURL,
+      imageURL,
+      name,
+      websiteURL,
+      rssName,
+      shoutouts: 0
```
cswenberg, line 24: if default shoutouts is zero, we should be able to set that in the entity definition. Better to define this intended behavior at the model level vs de-coupling it in the repo level.

tests/articlerepo.test.ts
```diff
@@ -0,0 +1,101 @@
...
+        shoutouts: 20
+      },
+      {
+        id: 'one',
+        title: 'Yi Hsin is still meeting with publications',
+        publicationID: '5fac75af7f68e1f1a4ae8b1b',
+        articleURL: 'www.cuappdev.com/yi',
+        imageURL: '',
+        date: new Date('11-05-2020'),
+        shoutouts: 10
```
cswenberg, line 86: might be useful to define a helper function for generating these objects to slim up the testing code and make it easier for generating entities in future tests

utils/createDummyData.ts
```diff
@@ -1,71 +1,68 @@
...
       title: 'We need better articles',
-      publication: 'advocate',
-      articleURL: 'www.ne',
-      imageURL: 'www.pic',
+      publicationID: '5fac75af7f68e1f1a4ae8b1b',
+      articleURL: 'www.newhat.com',
+      imageURL: 'www.pic.com',
       date: new Date('10-21-2020'),
-      likes: 0,
+      shoutouts: 0,
```
cswenberg, line 45: does setting all of these to the same `publicationID` not produce a crash? We are enforcing uniqueness so this should be a problem? It might be better to just generate new dummy uuids on the fly via the uuid function (whatever it is)

cswenberg (changes requested):
as your PR description notes, you should separate these into different PR's moving forward. Our principle should be smaller and sooner versus bigger and later! No PR reviewer ever said "wow I wish this PR was longer" 😉 . Also you should consider learning how to rebase so you can clean up your git history and commit naming for a cleaner history on main

src/repos/ArticleRepo.ts
```diff
@@ -0,0 +1,87 @@
...
+ * @param {string} id - string representing the unique Object Id of an article.
+ */
+const incrementShoutouts = async (id: string): Promise<Article> => {
+  const article = await ArticleModel.findById(new ObjectId(id));
+  article.shoutouts += 1;
+  const publication = await PublicationModel.findOne({ publicationID: article.publicationID });
+  publication.shoutouts += 1;
+  publication.save();
+  return article.save();
+};
```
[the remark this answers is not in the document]
cswenberg: just to take this line of thinking further: the number of articles and publishers scale differently. Because of this, article queries will take longer than the publisher ones, highlighting one of the problems of coupling their queries

src/repos/ArticleRepo.ts
```diff
@@ -0,0 +1,87 @@
...
+import { PublicationModel } from '../entities/Publication';
+import getRecentArticles from '../db/rss-parser';
+
+const getArticleById = async (id: string): Promise<Article> => {
+  return ArticleModel.findById(new ObjectId(id));
+};
+
+const getAllArticles = async (limit: number): Promise<Article[]> => {
+  return ArticleModel.find({}).limit(limit);
+};
```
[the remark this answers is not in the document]
tedim52: Good idea. I'll talk to frontend about the best number to put this at but will set it at 25 for now.

cswenberg:
Also to address your issue of people spamming shoutouts, you can make it a relational field that ties a user to an article/publisher so that we can verify each user can only shoutout once per. Not worth doing right now though and there may an easier way to do it.

cswenberg:
also @aliciaxw thoughts on setting a new standard where we create a `Changelog` section to the readme for high-level changes over time? Ik you theoretically can get the same insight by reading the commit history, but that's relatively cumbersome. For each PR we would make sure to add an entry that contains the date, author, and a couple bullets on the changes made.

utils/createDummyData.ts
tedim52: To address it crashing, I could be wrong but I think uniqueness is enforced upon entry into the database but in testing here, I've just creating Article typed objects and comparing equality of the objects then testing the compareTrendiness query. But if I 1. create field resolver for trendiness and 2. add a function to make the objects on the fly this shouldn't be a problem anyways.

src/repos/ArticleRepo.ts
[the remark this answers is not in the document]
cswenberg: also I would say modifying the shoutouts in this way is against OOP principles. Instead of manipulating the field from the outside, we should define a strict method within the entity for incrementing a shoutout. Also, I'm now noticing/guessing that a publication's shoutouts are the sum of all of its articles shoutouts? If this is true, we should make the publication `shoutout` field also a `Field Resolver` for carrying out that computation, freeing us of the need to update both entities (and ensuring our numbers never get out of sync). With these changes, this method should look more like
```
const incrementShoutouts = async (articleId: string): Promise<Article> => {
    const article = await ArticleModel.incrementShoutout(new ObjectId(articleId));
    return article.save()
}
```

publications.json
```diff
@@ -1,94 +1,84 @@
 {
    "publications":[
       {
-         "slug":"dailysun",
          "bio":"The Cornell Daily Sun is an independent daily newspaper published in Ithaca, New York by students at Cornell University and hired employees. The Sun features coverage of the university and its environs as well as stories from the Associated Press and UWIRE.",
-         "feed":"https://cornellsun.com/feed/",
-         "img":"https://twitter.com/cornellsun/photo",
+         "rssURL":"https://cornellsun.com/feed/",
+         "imageURL":"https://twitter.com/cornellsun/photo",
```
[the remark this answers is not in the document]
tedim52: Will note. Orko and I plan to create an assets folder in a future PR containing all publication static images like profile pic and info once we receive them from design.

aliciaxw:
> also @aliciaxw thoughts on setting a new standard where we create a `Changelog` section to the readme for high-level changes over time? Ik you theoretically can get the same insight by reading the commit history, but that's relatively cumbersome. For each PR we would make sure to add an entry that contains the date, author, and a couple bullets on the changes made.

I'd like to avoid manual documentation as much as possible a that's a lot of developer effort and is hard to maintain. If we can assert that commit messages be descriptive (good practice!) we can have the same result.

README.md
```diff
@@ -1 +1,53 @@
...
+
+`mongo`
+
+From here, you can interact with your database by typing `use <db_name>`.
+
+If you are using VSCode, we recommend installing the MongoDB extension to allow you to interact with documents in your database straight through the editor. More information on this [here](https://code.visualstudio.com/docs/azure/mongodb).
+
+## Environment variables
+
+We recommend using [`dotenv`](https://www.npmjs.com/package/dotenv). To set up, simply add a `.env` file to your source directory and add the following environment variables.
```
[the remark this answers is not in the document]
aliciaxw: yes, we should be using direnv and .envrc instead. we can add this as a separate issue and address in a separate pr

aliciaxw (approved):
amazing job!!!! i would address some of conner's comments (at least the simple fixes), but for the sake of moving quickly, let's push this as soon as we can and fix the larger scale architectural comments in later PRS :)

utils/createDummyData.ts
tedim52: Oh, I just realized all this means is that they're all from the articles from the same publication which is actually fine, doesn't affect the test, and is a realistic result.

src/repos/ArticleRepo.ts
```diff
@@ -0,0 +1,87 @@
...
+  const presentDate = new Date().getTime();
+  const a1Score = a1.shoutouts / (presentDate - a1.date.getTime());
+  const a2Score = a2.shoutouts / (presentDate - a2.date.getTime());
+  return a2Score - a1Score;
+};
+
+/**
+ * Computes and returns the trending articles in the database.
+ *
+ * Trendiness is computed by taking the number of total shoutouts an article
```
yuna878, line 37: I feel like this description should go on `compareTrendiness()`

yuna878 (approved):
Great first PR 🔥  Make sure to address comments above!

cswenberg:
@tedim52 once you create official issues for my comments, I will approve 😄

aliciaxw:
> @tedim52 once you create official issues for my comments, I will approve 😄

@cswenberg looks like the issues are made!

cswenberg (approved):

PR #12, opened 2020-12-10
Add assets and fix bios

orkosinha:
## Overview

This PR adds the correct image assets from [here](https://github.com/cuappdev/assets) for publications. Also fixes the bios and adds a new short biography for each publication.

## Changes Made

* Add the `slug`, `profileImageURL`, `backgroundImageURL`, and `bioShort` in `Publication.ts`.
* Alphabetize `publications.json`, `Publication.ts` and `Article.ts`.
* Update `publication.json` with all relevant data for publications.
* Populate the background and profile images for each publication with the correct image path from the assets repository.

## Test Coverage

Tested locally for each publication.


src/db/rss-parser.ts
```diff
@@ -39,9 +41,8 @@ const getRecentArticles = async (): Promise<Article[]> => {
             imageURL: parseImage(
               article['content:encoded'] ? article['content:encoded'] : article.content,
             ),
-            publicationID: nameToIDMap[publication.title]
-              ? nameToIDMap[publication.title]
-              : 'unknown',
+            publicationID: nameToIDMap[publication.title] || 'none',
+            publicationSlug: nameToSlugMap[publication.title] || 'none',
```
cswenberg, line 45: curious why we are using `'none'` strings vs just 'null' values?

src/entities/Publication.ts
```diff
@@ -8,31 +8,43 @@ export class Publication {
...
   @Property({ unique: true })
   rssName: string;
 
+  @Field()
+  @Property()
+  rssURL: string;
+
+  @Field()
+  @Property()
+  slug: string;
```
cswenberg, line 39: not sure if we need should place `unique: true` everywhere, but what are your thoughts on enforcing these as unique vs. not?

src/repos/PublicationRepo.ts
```diff
@@ -3,6 +3,10 @@ import { Publication, PublicationModel } from '../entities/Publication';
 
 import publicationsJSON from '../../publications.json';
 
+function getImageURL(slug: string, type: string): string {
+  return `${process.env.IMAGE_PATH}/${slug}/${type}.png`;
+}
+
```
cswenberg, line 9: can we add to the documentation in either `assets/volume` or in this repo that all images must be of type `.png`? I'm sure whoever is going to add images in the future will likely assume this much by looking at existing content, but we should still get in the happen of connecting niche business logic like this with some form of documentation for onboarding new backend devs once you hand this over to someone else

.dockerenvrc
```diff
@@ -1,3 +1,4 @@
 NODE_ENV='production'
 MONGO_ADDRESS='mongodb://mongo:27017/volume'
-SERVER_ADDRESS='http://volume-backend.cornellappdev.com'
\ No newline at end of file
+SERVER_ADDRESS='http://volume-backend.cornellappdev.com'
+IMAGE_PATH='https://raw.githubusercontent.com/cuappdev/assets/master/volume'
```
cswenberg: super nit-picky: For consistent variable naming, I think `IMAGE_ADDRESS` is more fitting?

src/repos/PublicationRepo.ts
```diff
@@ -13,16 +17,19 @@ const addPublicationsToDB = async (): Promise<void> => {
...
       Object.assign(new Publication(), {
         _id: new ObjectId().toString(),
+        backgroundImageURL: getImageURL(slug, 'background'),
         bio,
-        rssURL,
-        imageURL,
+        bioShort,
         name,
-        websiteURL,
+        profileImageURL: getImageURL(slug, 'profile'),
```
cswenberg, line 28: not sure if I like this more or less than
```
const { bio, bioShort, rssName, rssURL, name, slug, websiteURL } = publication;
const { backgroundImageURL, profileImageURL } = getImageURLs(slug) 
publications.push(
    Object.assign(new Publication(), {
        ...
        backgroundImageURL,
        ...
        profileImageURL,
        ...
```
thoughts on the two styles?
(this is not an important issue, I was just curious to think of another way of going about it)

cswenberg (approved):
lgtm, glad to see we are so close and cleaning up the codebase :)

src/entities/Publication.ts
orkosinha: Adding `unique: true` adds some extra overhead when adding a publication to the list. They should be unique, but since all of this content is generated by us with the hardcoded `publications.json` I don't think it's necessary. I did make the slug unique though, because that's just the definition of slugs and at least one field needs to be unique.

.dockerenvrc
orkosinha: Agreed and changed

src/repos/PublicationRepo.ts
orkosinha: Yeah, it looks a lot better moving them out. I used a Typescript tuple here though because otherwise the function signature with types becomes a little long.

src/entities/Publication.ts
tedim52: it says only the rssName is unique? not the slug

src/repos/PublicationRepo.ts
tedim52: also, more Volume specific but adding documentation on the type of image, 'profile' or 'background', we may add more assets for publications in the future depending on how UI changes?

src/repos/PublicationRepo.ts
tedim52: wouldn't using a tuple here constrict us to only return a profile and a background image? can't imagine a situation where it occurs rn but we might potentially add other publication assets in the future

tedim52 (approved):

PR #19, opened 2021-02-20
Add social urls to publications and two queries

tedim52:
## Changes Made

- Add getPublicationsByIDs query
- Add getArticlesByPublicationIDs
- Add field resolver to retrieve links to publications social platforms
- Remove since field on trending articles query

## Notes

1. Currently, social urls are queried straight from publications json rather than database. Ideally, they would be apart of the publication object in the database and the database would be queried but after hours of issues with mongo, I decided to go for a quick fix and come back to attempting this later. Opening an issue for this.
2. The since field was deemed unnecessary as articles are ranked by 'trendiness' which already accounts for time. Additionally, the since field makes it more difficult to query for an exact number of articles needed on a feed.

## Test Coverage
Postman test suite
[Volume Testing.postman_collection.zip](https://github.com/cuappdev/volume-backend/files/6019145/Volume.Testing.postman_collection.zip)

</details>


publications.json
```diff
@@ -7,15 +7,47 @@
...
+            {
+               "social":"linkedin",
+               "URL":"https://www.linkedin.com/company/bigredsportsnetwork"
+            },
+            {
+               "social":"twitter",
+               "URL":"https://twitter.com/CornellBRSN"
+            }
+         ],
          "websiteURL":"https://www.bigredsportsnetwork.org/"
```
alanna-zhou, line 51: is there a reason why this has trailing slash and the others don't

src/repos/PublicationRepo.ts
```diff
@@ -54,6 +55,16 @@ const getPublicationByID = async (id: string): Promise<Publication> => {
   return PublicationModel.findById(new ObjectId(id));
 };
 
+const getPublicationsByIDs = async (ids: string[]): Promise<Publication[]> => {
+  return Promise.all(ids.map((id) => PublicationModel.findById(new ObjectId(id)))).then(
+    (publications) => {
+      // Filter out all null values that were returned by ObjectIds not associated
```
alanna-zhou, line 61: i like the comment xd

src/repos/PublicationRepo.ts
```diff
@@ -110,12 +121,41 @@ const getNumArticles = async (publication: Publication): Promise<number> => {
   return pubArticles.length;
 };
 
+/**
+ * Maps list of dictionaries containing social platform with URL information to
+ * list of SocialURLTuple types.
+ */
+
```
alanna-zhou, line 128: remove white space

alanna-zhou (approved):
ur a beast

publications.json
orkosinha: That was def an accident

src/repos/ArticleRepo.ts
```diff
@@ -60,17 +60,12 @@ export const compareTrendiness = (a1: Article, a2: Article) => {
...
 const getTrendingArticles = async (
-  since: string,
   limit = Constants.DEFAULT_LIMIT,
 ): Promise<Article[]> => {
-  const articlesSinceDate = await ArticleModel.find({
-    date: { $gte: new Date(new Date(since).setHours(0, 0, 0)) },
-  }).exec();
-
-  return articlesSinceDate.sort(compareTrendiness).slice(0, limit);
+  const articles = await ArticleModel.find({}).exec();
```
orkosinha, line 67: Just curious as to why we need to find and then exec, does await not get the articles? 

I guess a better question is what does `exec()` do?

orkosinha (approved):
Love the test cases. Just one question and also address Alanna's comments. Awesome!

publications.json
tedim52: will fix

src/repos/ArticleRepo.ts
tedim52: https://mongoosejs.com/docs/promises.html#should-you-use-exec-with-await? that article should explain it, but i realized idk if all queries use .exec() so I'll review to make them consistent

tedim52: TLDR, .exec() just gives a better stack trace

PR #21, opened 2021-02-21
Add tests through Travis, Docker volume, landing page, and minor fixes

orkosinha:
## Overview

This PR does a few things.

1. Runs all of our Postman tests through Travis.
2. Adds a landing page so we don't have a "Cannot GET /". This page links to the GraphQL Playground. 
3. Fix for connecting to database before running the application.
4. Properly add in publications from the JSON rather than the jank way.
5. Use Docker volumes rather than restarting the MongoDB container on each deployment.
6. Took out the `utils` folder which doesn't add anything to this project anymore and moved some env variables to be constants.

#### Testing
I updated the test script in `package.json` so that the test script is all that is needed to start the server, run the tests, close the server. Also setup Travis to run the tests on every push so that we can get a quick indication of whether the code is valid. 

Also got rid of jest stuff.

#### Landing page
Added a landing page so we don't get hit by an awkward "Cannot GET /" page on the base route. This could've been a straight redirect to the `graphql` endpoint, but this way we can add other links that are important as needed.

Could also make the page look clean later on for some style points.

Screen Shot 2021-02-21 at 2 07 03 PM

#### Docker Volumes
Because I didn't really know how Docker works, every time the container was restarted we lost our article data. With Docker volumes our data should stay on restarts. Big oops on my part here.

#### Taking out `utils` and reducing env variables
These files aren't really necessary anymore. They were useful test scripts during development, but they don't add anything to this project and would likely not be used again. We could put these on gists and link to them in the README.
`IMAGE_ADDRESS` also isn't a configuration option, so it did not make sense to keep it around as an env variable, so this was taken out and moved to `Constants`.

Sorry for the large PR. I was working on a few of these fixes since break and didn't get around to making a PR until now. Very unfortunate, very sorry.


docker-compose.yml
```diff
@@ -2,22 +2,19 @@ version: "3.6"
 
 services:
   web:
```
alanna-zhou, line 4: make sure you update the server with this and see if you can deploy -- we have autodeployment, but you will still have to manually replace the `docker-compose.yml` on the server

docker-compose.yml
```diff
@@ -2,22 +2,19 @@ version: "3.6"
...
-    container_name: volume
+    container_name: "volume"
```
alanna-zhou, line 5: i don't recommend having container_name -- theres always weird errors once you deploy

alanna-zhou (approved):
nice

src/repos/PublicationRepo.ts
```diff
@@ -16,38 +14,30 @@ function getImageURLs(slug: string): [string, string] {
...
+      profileImageURL,
+      rssName,
+      rssURL,
+      slug,
+      websiteURL,
+    });
+
+    // Add or update the publication created from the JSON
+    pubDocUpdates.push(
+      PublicationModel.updateMany({ slug: { $eq: pubDoc.slug } }, pubDoc, { upsert: true }),
```
tedim52, line 36: great great great great fix

tedim52 (approved):
LGTM

PR #24, opened 2021-02-28
Add nsfw tag, article refresh cronjob, setup for Zapier webhook integration

tedim52:
## Overview
This PR just adds some final changes to the repo before launch and integration with prod server. Notably, a filter on article content with profanity. Additionally, I added a cronjob that will serve as the current solution for updating articles. However, as we are looking to use webhooks as a better design for updating articles, I added a mutation that will act as a trigger anytime a feed is updated to create a new article in the database. More information on this integration with Zapier in my next PR and in a tech talk on March 21st :)

## Changes Made
- Add nsfw tag
- Add createArticle mutation
- Add articleRefreshCronJob
- Update postman tests

## Test Coverage
Postman Test Suite


src/app.ts
```diff
@@ -33,6 +34,14 @@ const main = async () => {
 
   server.applyMiddleware({ app });
 
+  async function setupArticleRefreshCron() {
+    // Refresh articles every minute (testing purposes - will be 12 hours later)
+    cron.schedule('* * * * *', async () => {
```
orkosinha, line 39: Maybe instead of using a 3rd party service like Zapier just to ping a few RSS feeds, we can do a conditional GET request and then choose to parse every minute. I don't believe just doing a simple GET request for the last refresh date is too much overhead. 

It is too much overhead to be parsing each RSS feed every minute though (it's ~130 articles every minute that are read and put in / updated).

package.json
```diff
@@ -33,6 +35,7 @@
     "moment": "^2.29.1",
     "mongodb": "^3.6.2",
     "mongoose": "^5.10.9",
+    "node-cron": "^2.0.3",
     "newman": "^5.2.2",
```
orkosinha, line 39: nit but I believe this list is like partially alphabetized, can you alphabetize it? Will look into prettier taking care of this for us in the future

src/resolvers/ArticleResolver.ts
```diff
@@ -60,11 +60,27 @@ class ArticleResolver {
...
+  async nsfw(@Root() article: Article): Promise<boolean> {
+    return ArticleRepo.checkProfanity(article['doc'].title); //eslint-disable-line
+  }
+
   @Mutation((_returns) => [Article])
   async refresh() {
     return ArticleRepo.refreshFeed();
   }
 
+  @Mutation((_returns) => Article)
```
orkosinha, line 73: Potentially dangerous, if the same articleURL is provided on two separate mutations, this will crash.

tests/Volume Testing.postman_collection.json
```diff
@@ -126,7 +126,7 @@
 						"body": {
```
orkosinha, line 126: Is this name auto-generated from Postman? I changed it to just be lower case "volume" and that's how the Travis tests are setup. We can change the Travis tests or this file name, up to you idk what's best practice for something like this. I would prefer lower case if these aren't autogenerated.

tests/volume.postman_environment.json
```diff
@@ -1,54 +0,0 @@
-{
```
orkosinha, line 1: Don't we need this file for newman?

src/repos/ArticleRepo.ts
```diff
@@ -88,6 +89,27 @@ const refreshFeed = async (): Promise<Article[]> => {
...
+  content: string,
+): Promise<Article> => {
+  const articleObject = Object.assign(new Article(), {
+    articleURL,
+    date: new Date(pubDate),
+    imageURL: content,
+    publicationSlug: pub,
+    title,
+  });
+  const articleDoc = await ArticleModel.insertMany(articleObject, { ordered: false });
```
orkosinha, line 109: I believe this will crash if the articleURLs provided are the same. Also I think this is used for testing right? If so, keeping the test article in the DB may be a bit problematic.

src/resolvers/ArticleResolver.ts
orkosinha: I think this is a limitation of using newman tests. They aren't good for checking internal functionality, and something like `jest` would definitely be much more normal to use.

src/repos/ArticleRepo.ts
```diff
@@ -99,7 +121,19 @@ const incrementShoutouts = async (id: string): Promise<Article> => {
   return article.save();
 };
 
+/**
+ * Checks if an article's title contains profanity.
+ * @function
+ * @param {string} title - article title.
+ */
+const checkProfanity = async (title: string): Promise<boolean> => {
```
orkosinha, line 129: Love this

src/db/rss-parser.ts
```diff
@@ -46,7 +46,6 @@ const getRecentArticles = async (): Promise<Article[]> => {
       )
       .flat(),
   );
-
```
orkosinha, line 49: rip this one line in particular

src/resolvers/ArticleResolver.ts
```diff
@@ -60,11 +60,27 @@ class ArticleResolver {
     return article['_doc'].shoutouts / (presentDate - article['_doc'].date.getTime()); // eslint-disable-line
   }
 
+  @FieldResolver((_returns) => Number)
+  async nsfw(@Root() article: Article): Promise<boolean> {
+    return ArticleRepo.checkProfanity(article['doc'].title); //eslint-disable-line
```
orkosinha, line 65: I think we could do this in the parsing step. 

We don't store article text, but some RSS feeds do give us this data. This would not only make our NSFW flag more powerful with article content, but it would also prevent having to this categorization in every request.

src/resolvers/ArticleResolver.ts
```diff
@@ -60,11 +60,27 @@ class ArticleResolver {
     return article['_doc'].shoutouts / (presentDate - article['_doc'].date.getTime()); // eslint-disable-line
   }
 
+  @FieldResolver((_returns) => Number)
+  async nsfw(@Root() article: Article): Promise<boolean> {
```
alanna-zhou, line 64: `is_NSFW`?

alanna-zhou (approved):
dang you're a tank 🔥

orkosinha (changes requested):
Few comments here and there but the main meat of it is two things.

1. I don't think we should have a `createArticle` mutation. Even if you were to add collision checking with existing articleURLs (which may or may not be an issue), it gives people a really easy way to make random articles.
2. More question than comment, but why did you choose to do this as a field resolver?

Sub-thing: I'm still looking at some alternatives to using Zapier but if you want to use that, we totally can. 
Also, def try to get that green check from Travis.

tests/volume.postman_environment.json
tedim52: We do not actually, the environment information is contained within the Volume Testing.postman_collection.json. The environment file is just good documentation and can be used to pass off to other devs.

src/resolvers/ArticleResolver.ts
tedim52: Are you saying the XML in the feed could give us this information?

orkosinha: You can use the status code directly, if you receive 304 then you know the feed hasn't been refreshed and can not even parse any of the XML.

tests/Volume Testing.postman_collection.json
tedim52: They are auto generated which is why I prefer to keep them this way not best practice, but means i can easily export form Postman and paste in file directory after adding tests without having to rename the file. I adjusted corresponding commands in travis setup.

orkosinha (approved):

alanna-zhou (approved):
:O

PR #26, opened 2021-04-21
Update trending articles algorithm

tedim52:
## Overview

This PR alters how trending articles are calculated. The intended functionality is that trending articles are refreshed every 12 hours and are pulled from any random article published in the last 30 days.

## Changes Made
- getTrendingArticles

## Test Coverage

Testing using updated Postman script with a new test to check all articles returned have isTrending = true.
[Volume Testing.postman_collection.json.zip](https://github.com/cuappdev/volume-backend/files/6348101/Volume.Testing.postman_collection.json.zip)

</details>


tedim52:
I might merge this into an open the open branch of tedi/nsfw to make commits more consistent.

src/repos/ArticleRepo.ts
```diff
@@ -1,4 +1,5 @@
 import Filter from 'bad-words';
+import moment from 'moment';
```
orkosinha, line 2: Seems like moment is deprecated, can we not do something like `new Date().getDate() - 30` instead?

https://momentjs.com/docs/

orkosinha (approved):
lgtm, other than the use of moment which may be questionable based on the docs page.

src/repos/ArticleRepo.ts
tedim52: It is deprecated but its still a pretty popular and widely used library thats easy for date time usage as suggested @grebneerg what do you think?

src/repos/ArticleRepo.ts
```diff
@@ -48,30 +49,52 @@ const getArticlesAfterDate = async (since: string, limit = DEFAULT_LIMIT): Promi
...
 /**
  * Computes and returns the trending articles in the database.
  *
  * @function
  * @param {number} limit - number of articles to retrieve.
  */
 const getTrendingArticles = async (limit = DEFAULT_LIMIT): Promise<Article[]> => {
-  const articles = await ArticleModel.find({}).exec();
-  return articles.sort(compareTrendiness).slice(0, limit);
+  const articles = await ArticleModel.find({ isTrending: true }).exec();
```
alanna-zhou, line 59: i thought `find` only gave you the first result... https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

is that what you want?

alanna-zhou: also what is the exec for

src/repos/ArticleRepo.ts
```diff
@@ -48,30 +49,52 @@ const getArticlesAfterDate = async (since: string, limit = DEFAULT_LIMIT): Promi
...
+
+/**
+ * Refreshes trending articles.
+ */
+const refreshTrendingArticles = async (): Promise<Article[]> => {
+  // Set previous trending articles to not trending
+  const oldTrendingArticles = await ArticleModel.find({ isTrending: true }).exec();
+  oldTrendingArticles.forEach(async (a) => {
+    try {
+      const article = await ArticleModel.findById(new ObjectId(a._id)); //eslint-disable-line
```
alanna-zhou, line 71: `// eslint-disable-line`

src/repos/ArticleRepo.ts
```diff
@@ -48,30 +49,52 @@ const getArticlesAfterDate = async (since: string, limit = DEFAULT_LIMIT): Promi
...
+      article.isTrending = false;
+      await article.save();
+    } catch (err) {
```
alanna-zhou, line 74: dry `try` `catch`es are not recommended -- you should at least for one level catch for an expected error, or you should only just be checking if you can find the article with the id in the first place

src/repos/ArticleRepo.ts
```diff
@@ -48,30 +49,52 @@ const getArticlesAfterDate = async (since: string, limit = DEFAULT_LIMIT): Promi
...
+      console.log(err);
+    }
+  });
+
+  // Get new trending articles
+  const articles = await ArticleModel.aggregate()
+    // Get a sample of random articles
+    .sample(100)
+    // Get articles after 30 days ago
+    .match({ date: { $gte: moment().subtract(30, 'days').toDate() } });
```
alanna-zhou, line 84: you should save the `30` in a constant that you reference from a constants file or a settings file, maybe like `MAX_NUM_DAYS_OF_TRENDING_ARTICLES` or something like that

src/repos/ArticleRepo.ts
```diff
@@ -48,30 +49,52 @@ const getArticlesAfterDate = async (since: string, limit = DEFAULT_LIMIT): Promi
...
+
+  // Set isTrending to true
```
alanna-zhou, line 86: delete this comment

src/repos/ArticleRepo.ts
```diff
@@ -48,30 +49,52 @@ const getArticlesAfterDate = async (since: string, limit = DEFAULT_LIMIT): Promi
...
+  articles.forEach(async (a) => {
+    try {
+      const article = await ArticleModel.findById(new ObjectId(a._id)); //eslint-disable-line
```
alanna-zhou, line 89: `// eslint-disable-line`

alanna-zhou (changes requested):
some little things
very cool tedi!!

src/repos/ArticleRepo.ts
grebneerg: I think `ArticleModel` isn't an array. It's a different `find` implementation provided by mongoose/typegoose.

src/repos/ArticleRepo.ts
grebneerg: I was unaware that moment has been deprecated. My instinct would be not to use it, especially given that they aren't going to add immutability and recommend libraries that do. [Luxon](https://moment.github.io/luxon/) seems to be the literal successor, hosted under the same organization on github, and looks very nice, but [day.js](https://day.js.org/) looks pretty good as well. I would still stay away from using built in JS dates directly, since `Date` objects are mutable and parsing and timezone handling really isn't great.

src/repos/ArticleRepo.ts
tedim52: +1 @grebneerg . Also to answer your second question: https://mongoosejs.com/docs/promises.html#should-you-use-exec-with-await?

alanna-zhou (approved):
youre a beast 💯

PR #29, opened 2021-05-10
Onboarding Cornell Creatives, The 180, Cornell Diplomat

tedim52:
## Overview

Adds more publications to the platform. Also adds a catch to ensure that an error is not brought when a publication ID is invalid, rather an empty array is returned if the publication does not exist. This will prevent frontend from freezing if publication IDs get swapped for whatever reason.

## Changes Made

-publication.json 
-add check to getArticlesByPublicationID

## Test Coverage
Postman tests.


alanna-zhou (approved):
seems to be just scripts stuff so cant rly complain

camfruss (approved):
For being named the Cornell Creatives, they need a more creative bio. Can Zain fix that? 

Is the indentation in `tests/Volume Testing.postman_collection.json` just a GitHub formatting issue?

tedim52:
@camfruss yes to second. Also, they give us whatever bio they want so up to Publication what they want.

PR #32, opened 2021-05-24
Add yes input to docker system prune -a in auto deployment

alanna-zhou:
Turns out we haven't been pruning automatically because Docker asks for confirmation on whether or not you'd like to nuke all the images not being used -- so I'm just piping in the `yes`! 

ez fix ;)


tedim52:
Alanna does it again!! you're a beast

tedim52 (approved):

PR #33, opened 2021-10-27
Implement User entity, UserRepo, and UserResolver

tedim52:
## Overview

This pull request adds a notion of a "pseudo" user to volume backend. The backend associates a unique id with a deviceToken and device type for the purposes of sending notifications(PR coming soon) and tracking followed publications on the backend. 

## Changes Made
- Add User entity
- Add UserRepo with functions to handle user logic
- Add UserResolver with GraphQL mutations

## Test Coverage

Postman tests coming soon!

Manually tested with GraphQL
Screen Shot 2021-10-27 at 4 35 03 PM

## Next Steps
- Finalize notifications implementation (using logic for Users)
- Create logic to retrieve articles with uuids and followedPublications(allows frontend to stop caching pubIDs)


src/repos/UserRepo.ts
```diff
@@ -0,0 +1,74 @@
...
+
+const getUserByUUID = async (uuid: string): Promise<User> => {
+  return UserModel.findOne({ uuid });
+};
+
+/**
+ * Return all users who follow a publication.
+ */
+const getUsersFollowingPublication = async (pubID: PublicationID): Promise<User[]> => {
+  const users = await UserModel.find(); // linear scan on DB, inefficient
```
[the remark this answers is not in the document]
noah-solomon: yeah this looks like a fat database query but probably not too many ways to improve other than what you mentioned

src/entities/User.ts
```diff
@@ -0,0 +1,30 @@
...
+  @Property({ required: true, type: () => PublicationID, default: [] })
+  followedPublications: mongoose.Types.DocumentArray<DocumentType<PublicationID>>;
+
+  @Field()
+  @Property()
+  deviceToken: string;
+
+  @Field()
+  @Property()
+  notification: string; // ANDROID, IOS
```
noah-solomon, line 27: not sure if this is nit picky, maybe this should be `deviceType` or something? notification seems a bit vague

noah-solomon (dismissed):
LGTM - still learning about volume backend & typescript in general but everything looks good 🤝

src/repos/UserRepo.ts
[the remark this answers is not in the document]
tedim52: fixed

src/entities/Article.ts
```diff
@@ -37,6 +37,10 @@ export class Article {
   @Field()
   @Property({ default: false })
   isTrending: boolean;
+
+  @Field()
+  @Property({ default: false })
+  notificationSent: boolean;
```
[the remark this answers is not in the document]
tedim52: fixed

src/entities/User.ts
tedim52: agreed

src/repos/UserRepo.ts
```diff
@@ -58,10 +58,8 @@ const getUserByUUID = async (uuid: string): Promise<User> => {
  * Return all users who follow a publication.
  */
 const getUsersFollowingPublication = async (pubID: PublicationID): Promise<User[]> => {
-  const users = await UserModel.find(); // linear scan on DB, inefficient
-  users.filter((u) => {
-    u.followedPublications.map((id) => id.id).includes(pubID.id);
-  });
+  const users = await UserModel.find({ followedPublications: pubID }); // linear scan on DB, inefficient
+  console.log(users);
```
noah-solomon, line 62: remove?

noah-solomon (approved):
Looks good to me

src/repos/UserRepo.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  return UserModel.create(newUser);
+};
+
+/**
+ * Adds publication of pubID to user represented by uuid's followedPublications.
+ */
+const followPublication = async (uuid: string, pubID: string): Promise<User> => {
+  const user = await UserModel.findOne({ uuid });
+  user.followedPublications.push(Object.assign(new PublicationID(), { id: pubID }));
```
orkosinha, line 34: I think it's worth checking if the pubID is valid before creating it?

src/repos/UserRepo.ts
```diff
@@ -0,0 +1,72 @@
...
+ */
+const unfollowPublication = async (uuid: string, pubID: string): Promise<User> => {
+  const user = await UserModel.findOne({ uuid });
+  const pubIDs = user.followedPublications.map((pubIDObject) => {
+    return pubIDObject.id;
+  });
+  if (pubIDs.indexOf(pubID) === -1) {
+    return user;
+  }
+  user.followedPublications.splice(pubIDs.indexOf(pubID), 1);
```
orkosinha, line 49: Can we do a filter rather than splice. This way the logic can be something like `user.followedPublications = user.followedPublications.filter(pubIDObj => pubIDObj.id !== pubID);`. No need to unwrap each object, store it and splice it out.

orkosinha (changes requested):
lgtm, few comments regarding following publications and schema design. The schema design one is pretty minor comment wise but major change front-end wise, so we can skip it over.

src/repos/UserRepo.ts
orkosinha: Actually, didn't realize we use `indexOf` twice. We should definitely make this change

orkosinha: Actually, didn't realize we use `indexOf` twice. We should definitely make this change

src/repos/UserRepo.ts
tedim52: I think it's okay not to check if the pubID is valid for the following reasons: 1. PubIDs are only ever made on Publication object creation which happens once during bootstrap 2. All queries that we have that use pubIDs to query articles just return an empty array if the pubID is invalid so there already exists a check to avoid consequences from invalid pubIDs. I don't think it's a bad idea to add an extra layer of protection against invalid pubIDs, but I don't know how to check validity without some how querying the DB or scanning the publication json for a list of pubIDs and I'd like to avoid that expense if possible.

src/entities/User.ts
```diff
@@ -0,0 +1,30 @@
...
+  @Property({ required: true, type: () => PublicationID, default: [] })
+  followedPublications: mongoose.Types.DocumentArray<DocumentType<PublicationID>>;
+
+  @Field()
+  @Property()
+  deviceToken: string;
+
+  @Field()
+  @Property()
+  deviceType: string; // ANDROID, IOS
```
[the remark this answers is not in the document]
tedim52: I think the extra storage space is negligible. I also think its more intuitive to distinguish using some keyword. If we decided to add logic to support desktop notifications down the line, we wouldn't be able to use a boolean. An "deviceType" enum type is probably the best structure to use but I think string is fine for now to avoid issues with GraphQL types.

src/repos/UserRepo.ts
tedim52: Agreed but I couldn't get that to work : (

orkosinha (approved):
LGTM!!!!

🚀🚀🚀

noah-solomon (approved):
looks spectacular

kidzegeye (approved):
LGTM

PR #36, opened 2021-11-12
De-ship refreshing feed

orkosinha:
## Overview
Removes refresh article methods.

## Changes Made

Removes initial refreshing and cron-job setup in `app.ts`

Removes corresponding mutation and repo function to refresh feeds. This method is not used by front-end so we do not need to support it for compatibility with legacy versions of the app.

This functionality is implemented in volume-microservice.


tedim52 (approved):
LGTM.

PR #38, opened 2021-11-13
Modify deployment workflow to acknowledge microservice

orkosinha:
## Overview
Change deployment script for volume-backend to deploy with acknowledging our microservice.

## Changes Made

Modify deployment script to track latest deployment tags inside of a file stored on the server.

Tracks image tags in a `tags` file located at home directory of server. This tracks the most recently used tags for volume-microservice, and will deploy that most recently used version along with the updated volume-backend.

Had to add `--with-registry-auth` to the deploy command based on [this](https://stackoverflow.com/questions/47470115/docker-stack-deploy-results-in-no-such-image-error).

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -31,7 +31,6 @@ const main = async () => {
 
   server.applyMiddleware({ app });
 
-
   async function setupTrendingArticleRefreshCron() {
     // Refresh trending articles 12 hours
     cron.schedule('0 */12 * * *', async () => {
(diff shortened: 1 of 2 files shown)


src/app.ts
```diff
@@ -31,7 +31,6 @@ const main = async () => {
 
   server.applyMiddleware({ app });
 
-
```
orkosinha, line 34: From prettier

tedim52 (approved):
LGTM.

PR #42, opened 2021-11-29
Fix getArticlesByPublicationID to support new schema

orkosinha:
## Overview

Quick fix to allow getArticlesByPublicationID to function as intended. Should be deprecated in migrating publication IDs to slugs.

## Changes Made

Fixes the query in GraphQL by redefining the function. Now it retrieves the publication from the database and queries the articles collection by the slug of the publication.


kidzegeye (commented):
looks good

PR #43, opened 2021-11-29
Implement Weekly Debrief

kidzegeye:
## Overview

Implemented the Weekly Debrief feature.

## Changes Made

- Created the WeeklyDebrief class
- Created readArticle and bookmarkArticle mutations for the User class. 
- Created the createWeeklyDebriefs function that creates the weekly debriefs for all users.
- Removed unique constraint from the slug field in Publication.ts because, as multiple users could not be created with the constraint active.

## Test Coverage

Tested each mutation manually. Tested creating weekly debriefs with a test mutation that ran the createWeeklyDebriefs function. 

## Next Steps

Need to create a cron job for the createWeeklyDebriefs function that creates weekly debriefs weekly.


src/repos/UserRepo.ts
```diff
@@ -20,13 +23,22 @@ const createUser = async (
     const followedPublications = followedPublicationsIDs.map((id) => {
       return Object.assign(new PublicationID(), { id });
     });
-
     const uuid = uuidv4();
+
+    const weeklyDebrief = Object.assign(new WeeklyDebrief(), {
+      uuid,
+      createdAt: new Date(),
+      expirationDate: new Date('December 17, 1995 03:24:00'),
```
orkosinha, line 31: I've always wondered what the significance of this date is

src/entities/Article.ts
```diff
@@ -8,7 +8,7 @@ export class Article {
   id: string;
 
   @Field()
-  @Property({ unique: true })
+  @Property()
```
orkosinha, line 11: Why can't this be `unique`

src/entities/WeeklyDebrief.ts
```diff
@@ -0,0 +1,42 @@
...
+  @Field()
+  @Property({ default: 0 })
+  numBookmarkedArticles?: number;
+
+  @Field()
+  @Property({ default: 0 })
+  numReadArticles?: number;
+
+  @Field((type) => [Article])
+  @Property({ required: true, type: () => Article, default: [] })
```
orkosinha, line 36: `@Property({ required: true, type: [Article] })` does this not work?

src/entities/WeeklyDebrief.ts
```diff
@@ -0,0 +1,42 @@
...
+  readArticles: mongoose.Types.DocumentArray<DocumentType<Article>>;
```
orkosinha, line 37: And then `readArticles: Article[]`

src/entities/WeeklyDebrief.ts
```diff
@@ -0,0 +1,42 @@
...
+import { Article } from './Article';
+
+@ObjectType({ description: 'The Weekly Debrief Model' })
+export default class WeeklyDebrief {
+  @Field(() => ID)
+  id: string;
+
+  @Field()
+  @Property()
+  uuid: string; // uuid of user associated with this WeeklyDebrief
```
orkosinha, line 13: Does this show up as a comment in the Apollo GraphQL interface? If not, try to make it comment that does, it may be a param in the decorator, not sure though

src/repos/UserRepo.ts
```diff
@@ -75,8 +87,49 @@ const getUsersFollowingPublication = async (pubSlug: string): Promise<User[]> =>
...
+    user.save();
+  }
+  return user;
+};
+
+/**
+ * Increment number of bookmarks in user's numBookmarkedArticles
+ */
+const incrementBookmarks = async (uuid: string): Promise<User> => {
+  const user = await UserModel.findOne({ uuid });
```
orkosinha, line 120: You could use `findOneAndUpdate({'uuid': uuid}, {$inc: {'numBookmarkedArticles`: 1}})`. I believe it'd be faster but not exactly sure on the syntax bc mongoose docs are bad. [Here](https://mongoosejs.com/docs/api.html#model_Model.findOneAndUpdate) it is though.

orkosinha (dismissed):
Great work so far!!

I'll take a deeper look into this soon but here are some initial comments

src/repos/UserRepo.ts
tedim52: +1

tedim52: should just be curr date + 7

src/entities/WeeklyDebrief.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  @Field()
+  @Property()
+  createdAt: Date;
```
tedim52, line 17: lets change this to `creationDate` to make it consistent with expirationDate.

src/entities/Article.ts
tedim52: I believe it caused errors when trying to embed publication object within the article but not 100% sure. I remember we had a conversation about this and deemed it no longer necessary for it to be true for some reason...

src/entities/WeeklyDebrief.ts
tedim52: +1

src/entities/WeeklyDebrief.ts
tedim52: I believe I've attempted this before and it didn't work. I think Kidus copied this from my implementation of followedPublications and `@Property({ required: true, type: () => Publication, default: [] })
  followedPublications: mongoose.Types.DocumentArray<DocumentType<Publication>>;` was the only way I could get a list of Publication objects after looking through several github issues. Not sure why though to be honest.

src/repos/UserRepo.ts
```diff
@@ -75,8 +87,49 @@ const getUsersFollowingPublication = async (pubSlug: string): Promise<User[]> =>
...
+/**
+ * Increment shoutouts in user's numShoutouts
+ */
+const incrementShoutouts = async (uuid: string): Promise<User> => {
+  const user = await UserModel.findOne({ uuid });
+  if (user !== null) {
+    user.numShoutouts += 1;
+    user.save();
+  }
+  return user;
```
tedim52, line 113: if the user is null, doesn't this show up as a graphql error? or is it just an empty object?

tedim52 (changes requested):
Great PR! Good work. Just address some of the changes and we should be good to go.

src/entities/WeeklyDebrief.ts
kidzegeye: Is there a way to do this using TypeGraphQL? I've been looking around in their docs but I'm not finding anything.

src/repos/UserRepo.ts
kidzegeye: Fixed! Should return null now in that case.

src/repos/UserRepo.ts
kidzegeye: I've tested out both solutions and while both work, it seems that both take around 9-15ms to run per bookmark. A con I found of ```findOneAndUpdate``` was that I found were that I had to disable ```findAndModify``` in order to use it because it is deprecated otherwise. I'm having trouble finding what this changes, so I'm kinda worried about using this method.

src/repos/UserRepo.ts
kidzegeye: What behavior should happen if the user is created in the middle of the week?

kidzegeye: This is the weeklydebrief that is created for new users, so shouldn't it start out as expired?

tedim52: You are correct. I rescind my earlier comment. Instead, make this a constant such as `IMMEDIATELY_EXPIRE_DATE` or something. It's unclear what a hardcoded date is - future developers reading the code may not understand the intention of it. Changing it to a constant with an accurate name will display the intention of setting it to a date that immediately expires the weekly debrief.

src/entities/Publication.ts
```diff
@@ -35,7 +35,7 @@ export class Publication {
   rssURL: string;
 
   @Field()
-  @Property({ unique: true })
+  @Property() // Couldn't create multiple users with unique constraint active
```
tedim52, line 38: I would suggest removing this comment and adding this information to the "Changes Made" in the README. The comment doesn't make sense in the context of the code, it only makes sense when someone is comparing the previous version of this line of code, to the current. Therefore, it doesn't belong in code, rather version control.

kidzegeye: Done

src/repos/WeeklyDebriefRepo.ts
```diff
@@ -0,0 +1,55 @@
+import { DocumentType } from '@typegoose/typegoose';
+import { User, UserModel } from '../entities/User';
+import WeeklyDebrief from '../entities/WeeklyDebrief';
+import { ArticleModel } from '../entities/Article';
+
+const getExpirationDate = (creationDate: Date): Date => {
+  creationDate.setDate(creationDate.getDate() + 7);
```
dishanest, line 7: what happens when `creationDate.getDate()` is 29? does `setDate` automatically roll those extra days over to the next month?

dishanest (approved):
lgtm! just a few questions to look out for

src/repos/UserRepo.ts
```diff
@@ -24,11 +25,12 @@ const createUser = async (
...
     const uuid = uuidv4();
-
+    const creationDate = new Date();
+    const expirationDate = new Date(IMMEDIATELY_EXPIRE_DATE);
     const weeklyDebrief = Object.assign(new WeeklyDebrief(), {
       uuid,
-      creationDate: new Date(),
-      expirationDate: new Date('December 17, 1995 03:24:00'),
+      creationDate,
+      expirationDate,
```
snajima, line 33: I'm not too familiar with the codebase, but what is the point of making this "invalid" `WeeklyDebrief` object? Can you not define the User model's `weeklyDebrief` field as `weeklyDebrief: WeeklyDebrief | undefined;`, for example? Sounds like a super jank workaround that could confuse future members working on this...

src/repos/UserRepo.ts
```diff
@@ -20,13 +24,24 @@ const createUser = async (
...
+
     const newUser = Object.assign(new User(), {
       uuid,
       deviceToken,
       followedPublications,
       deviceType,
+      weeklyDebrief,
+      readArticles: [],
     });
     return UserModel.create(newUser);
```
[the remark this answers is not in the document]
snajima: This feels important to address in comments/documentation too

src/repos/WeeklyDebriefRepo.ts
tedim52: A different note, on this same line of code I didn't think about, didn't we decide there was a uniform day in which all weekly debrief objects were created? In which case shouldn't they all expire at the same time too? I could be forgetting the logic we decided on, but if this is the case, it shouldn't be `getDate() + 7` but rather setting to the next upcoming "sunday" or whatever time weekly debriefs are created on.

src/common/constants.ts
```diff
@@ -3,3 +3,4 @@ export const IMAGE_ADDRESS = 'https://raw.githubusercontent.com/cuappdev/assets/
 export const MAX_NUM_DAYS_OF_TRENDING_ARTICLES = 30;
 export const IOS = 'IOS';
 export const ANDROID = 'ANDROID';
+export const IMMEDIATELY_EXPIRE_DATE = 'December 17, 1995 03:24:00';
```
chalo2000, line 6: what is this constant for? I'm not sure what is 'immediately expiring'

src/repos/ArticleRepo.ts
```diff
@@ -97,8 +97,11 @@ const refreshTrendingArticles = async (): Promise<Article[]> => {
  */
 const incrementShoutouts = async (id: string): Promise<Article> => {
   const article = await ArticleModel.findById(new ObjectId(id));
-  article.shoutouts += 1;
-  return article.save();
+  if (article !== null) {
```
chalo2000, line 100: you can probably just do `if (article)`

src/repos/UserRepo.ts
chalo2000: I see what IMMEDIATELY_EXPIRE_DATE is for now; ideally we can leave `weeklyDebrief` as `undefined` as @snajima mentioned until the weekly debriefs are generated for the first time in which case we can have a valid creation and expiration date.

src/repos/UserRepo.ts
```diff
@@ -38,6 +53,9 @@ const createUser = async (
  */
 const followPublication = async (uuid: string, pubID: string): Promise<User> => {
   const user = await UserModel.findOne({ uuid });
+  if (!user) {
```
chalo2000, line 56: can we swap this with checking
```
if (user) {
   user.followedPublications.push(Object.assign(new PublicationID(), { id: pubID }));
   return user.save();
}
return user;
```

so that we can keep consistent with the other ifs in the codebase. Do this for lines 68-70 as well.

src/repos/UserRepo.ts
```diff
@@ -75,8 +96,54 @@ const getUsersFollowingPublication = async (pubSlug: string): Promise<User[]> =>
...
+  user.save();
+  return user;
+};
+
+/**
+ * Increment shoutouts in user's numShoutouts
+ */
+const incrementShoutouts = async (uuid: string): Promise<User> => {
+  const user = await UserModel.findOne({ uuid });
+  if (user !== null) {
```
chalo2000, line 123: ditto with simply checking `if (user)`

chalo2000 (changes requested):
some changes to be made but pretty good :)

src/repos/UserRepo.ts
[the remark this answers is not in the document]
kidzegeye: Currently you can't create a user when another user with the same deviceToken already exists. I can document this behavior as well.

src/repos/WeeklyDebriefRepo.ts
kidzegeye: `setDate` does automatically go to the next month in that case. Also what specific day and time should we have the universal expiration date happen?

src/common/constants.ts
kidzegeye: Since new accounts can be made mid week, their weekly debriefs would not really be valid. But I think it will be better to make new users' weekly debriefs be `undefined` like Shungo proposed instead.

src/repos/WeeklyDebriefRepo.ts
kidzegeye: So I've been looking around for a solution using the Date() class and I can't really find any way to do it with a specific date. Wouldn'rt having the `getDate() + 7`  portion run on the cronjob be sufficient? Currently new users have a null weekly debrief, so all expiration dates should be uniform

snajima: I'm not sure how we'd handle timezones, so ignoring that for now, I was able to find [this](https://www.w3schools.com/jsref/jsref_getday.asp) to get the day of the week. If I'm understanding @tedim52 correctly and we want weekly debriefs to update on Sundays, something like this:
`const currentTime = new Date();
const daysUntilSunday = 7 - currentTime.getDay();
creationDate.setDate(currentTime.getDate() + daysUntilSunday);
creationDate.setHours(9);
creationDate.setMinutes(0);`
could work (it would set the creationDate to be 9am on the next Sunday). Unsure if that's what y'all are looking for though.

tedim52: This is resolved. The cronjob runs once a week, so `getDate() + 7` would set expirationDate to the same time for all cronjobs as desired.

src/resolvers/ArticleResolver.ts
```diff
@@ -2,65 +2,69 @@ import { Resolver, Mutation, Arg, Query, FieldResolver, Root } from 'type-graphq
 import { Article } from '../entities/Article';
 import ArticleRepo from '../repos/ArticleRepo';
 import { DEFAULT_LIMIT } from '../common/constants';
+import UserRepo from '../repos/UserRepo';
 
 @Resolver((_of) => Article)
 class ArticleResolver {
-  @Query((_returns) => Article, { nullable: true })
+  @Query((_returns) => Article, { nullable: true, description: 'Returns a single <Article> via the given <id>' })
```
dishanest, line 9: Do these descriptions show up in the Docs on the GraphQL playground?

dishanest (dismissed):
Great! Just one question.

src/resolvers/ArticleResolver.ts
katesliang: ^ bump

katesliang (dismissed):
LGTM 💯

src/repos/UserRepo.ts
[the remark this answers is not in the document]
chalo2000: Looks like this is already documented on lines 16-17

src/repos/UserRepo.ts
```diff
@@ -75,8 +84,54 @@ const getUsersFollowingPublication = async (pubSlug: string): Promise<User[]> =>
...
 };
 
+/**
+ * Add article to a user's readArticles
+ */
+const appendReadArticle = async (uuid: string, articleID: string): Promise<User> => {
+  const article = await ArticleRepo.getArticleByID(articleID);
+  const user = await UserModel.findOne({ uuid });
+  const checkDuplicates = (prev: boolean, cur: Article) => prev || cur.id === articleID;
+  if (!user) {
```
chalo2000, line 94: probably want to check this before instantiating the checkDuplicates so we instantiate variables just before they should be used

src/repos/UserRepo.ts
```diff
@@ -75,8 +84,54 @@ const getUsersFollowingPublication = async (pubSlug: string): Promise<User[]> =>
...
+    return user;
+  }
+  if (article) {
```
chalo2000, line 97: same here, we'd instantiate the article just before this instead of at the beginning of the function where it's just hanging around not doing much till it's used here

src/repos/WeeklyDebriefRepo.ts
```diff
@@ -0,0 +1,55 @@
...
+  const articleAggregate = ArticleModel.aggregate();
+  const { uuid } = user;
+  const weeklyDebrief = Object.assign(new WeeklyDebrief(), {
+    uuid,
+    creationDate,
+    expirationDate,
+    numShoutouts: user.numShoutouts,
+    numBookmarkedArticles: user.numBookmarkedArticles,
+    readArticles: user.readArticles.slice(0, 2),
+    numReadArticles: user.readArticles.length,
```
chalo2000, line 25: Why are only the first two articles of readArticles stored in readArticles? It seems a little strange to have a readArticles and numReadArticles - why store 2 fields when just storing `user.readArticles` and performing the `.slice` and `.length` operator later would keep the model simpler and make us store less information.

src/entities/WeeklyDebrief.ts
```diff
@@ -0,0 +1,42 @@
...
+  @Property({ default: 0 })
+  numReadArticles?: number;
+
+  @Field((type) => [Article])
+  @Property({ required: true, type: () => Article, default: [] })
+  readArticles: mongoose.Types.DocumentArray<DocumentType<Article>>;
+
+  @Field((type) => [Article])
+  @Property({ required: true, type: () => Article, default: [] })
+  randomArticles: mongoose.Types.DocumentArray<DocumentType<Article>>;
```
chalo2000, line 41: Maybe some documentation on what randomArticles is? I'm assuming that we randomize once and then store those two random picks but I also don't see why not do something like using a seed that we can guarantee getting the same random articles back (for example, the seed used to get the random articles for a user will always be their ID or, to make it more random, will be the ID of the weekly debrief).

src/resolvers/PublicationResolver.ts
```diff
@@ -6,42 +6,42 @@ import { Social } from '../common/types';
...
 
-  @FieldResolver((_returns) => Article, { nullable: true })
+  @FieldResolver((_returns) => Article, { nullable: true, description: 'The most recent <Article> of a <Publication>' })
   async mostRecentArticle(@Root() publication: Publication): Promise<Article> {
     return PublicationRepo.getMostRecentArticle(publication);
   }
 
-  @FieldResolver((_returns) => Number)
+  @FieldResolver((_returns) => Number, {description: 'The total shoutouts of a <Publication\'s> <Articles>'})
   async shoutouts(@Root() publication: Publication): Promise<number> {
```
chalo2000, line 35: numShoutouts to stay consistent with the naming of numArticles

src/resolvers/UserResolver.ts
```diff
@@ -1,10 +1,18 @@
...
 
 @Resolver((_of) => User)
 class UserResolver {
-  @Mutation((_returns) => User)
+  @Query((_returns) => User, { nullable: true, description: 'Returns a single [User] via a given [uuid]'})
+  async getUser(@Arg('uuid') uuid: string) {
+    return UserRepo.getUserByUUID(uuid);
+  }
+
+  @Mutation((_returns) => User, {description: `Creates a single <User> via given <deviceToken>, <followedPulbications>,
```
chalo2000, line 14: `<followedPublications>` misspelled

src/resolvers/UserResolver.ts
```diff
@@ -1,10 +1,18 @@
...
+   and <deviceType>. Will not create a user if one with the same <deviceToken> already exists.`})
```
chalo2000, line 15: Should be more explicit then 'will not create a user' (then what?). Creates a single user given unique fields, otherwise does nothing.

src/resolvers/UserResolver.ts
```diff
@@ -14,17 +22,35 @@ class UserResolver {
...
+    return user;
+  }
+
+  @Mutation((_returns) => User, { nullable: true, description: 'Increments the number of bookmarks for the <User> given by <uuid>'  })
+  async bookmarkArticle(@Arg('uuid') uuid: string) {
+    const user = await UserRepo.incrementBookmarks(uuid);
+    return user;
+  }
+
+  @Mutation((_returns) => [User], {description: 'Creates and returns the Weekly Debrief for all users'})
```
chalo2000, line 49: 'Creates and returns...' but in the other create functions we did not talk about returning. Stay consistent with the documentation so that there are no ambiguities (oh this one creates, but does it return?)

chalo2000 (changes requested):
There are lot of style, naming and documentation issues that I'd like to be corrected before I can approve this.

src/resolvers/ArticleResolver.ts
kidzegeye: Yes it does

src/repos/WeeklyDebriefRepo.ts
dishanest: On the [weekly debrief UI](https://www.figma.com/file/vMfqUmHwm9VHGChOjPQFlf/?node-id=3%3A2), we only show two of the articles that a user read over the course of the past week for sharing or revisiting. The `numReadArticles` field shows the user the summarized statistic of _how many_ articles they read over the last week.

src/resolvers/PublicationResolver.ts
dishanest: frontend is swamped rn, let's not make this change until later because it'll break our API

src/entities/WeeklyDebrief.ts
dishanest: The MVP avoids dealing with this kind of detail. We still don't even know if users care about random articles yet.

chalo2000 (approved):
LGTM

src/repos/UserRepo.ts
```diff
@@ -1,6 +1,8 @@
 import { v4 as uuidv4 } from 'uuid';
 import { User, UserModel } from '../entities/User';
 import PublicationRepo from './PublicationRepo';
+import ArticleRepo from './ArticleRepo';
+import { Article } from '../entities/Article';
```
alanna-zhou, line 5: alphabetize imports

src/repos/UserRepo.ts
```diff
@@ -16,17 +18,18 @@ const createUser = async (
 
   // if no user, create a new one
   if (!users[0]) {
```
alanna-zhou, line 18: how does this not give an error? what if it's an empty list.. shouldn't we be checking for length

src/repos/UserRepo.ts
```diff
@@ -75,8 +84,54 @@ const getUsersFollowingPublication = async (pubSlug: string): Promise<User[]> =>
...
+  if (user) {
+    user.numBookmarkedArticles += 1;
+    user.save();
+  }
+  return user;
+};
+
 export default {
   createUser,
+  incrementShoutouts,
```
alanna-zhou, line 132: exports should be alphabetized too

src/repos/WeeklyDebriefRepo.ts
alanna-zhou, line 7: 7 should be a constant in a constants file for num days in week or sth

src/repos/WeeklyDebriefRepo.ts
```diff
@@ -0,0 +1,55 @@
...
+): Promise<User> => {
+  const articleAggregate = ArticleModel.aggregate();
+  const { uuid } = user;
+  const weeklyDebrief = Object.assign(new WeeklyDebrief(), {
+    uuid,
+    creationDate,
+    expirationDate,
+    numShoutouts: user.numShoutouts,
+    numBookmarkedArticles: user.numBookmarkedArticles,
+    readArticles: user.readArticles.slice(0, 2),
```
alanna-zhou, line 24: what if the length is not more than 2.... this would give an error?

alanna-zhou (dismissed):
only some cosmetic changes

src/repos/UserRepo.ts
kidzegeye: In javascript it returns undefined in that case, which is falsy

src/repos/WeeklyDebriefRepo.ts
kidzegeye: It would return the entire array
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice

src/common/constants.ts
```diff
@@ -3,3 +3,4 @@ export const IMAGE_ADDRESS = 'https://raw.githubusercontent.com/cuappdev/assets/
 export const MAX_NUM_DAYS_OF_TRENDING_ARTICLES = 30;
 export const IOS = 'IOS';
 export const ANDROID = 'ANDROID';
+export const DAYS_IN_WEEK = 7;
```
alanna-zhou, line 6: constants should be alphabetized or organized in a way that makes sense

alanna-zhou (approved):
nice! constants files are always nice :)

dishanest (approved):
Lookin good!

src/repos/WeeklyDebriefRepo.ts
```diff
@@ -0,0 +1,56 @@
...
```
alanna-zhou, line 25: you're sure this wont give an error right..?

src/resolvers/UserResolver.ts
```diff
@@ -14,17 +26,49 @@ class UserResolver {
...
+  async bookmarkArticle(@Arg('uuid') uuid: string) {
+    const user = await UserRepo.incrementBookmarks(uuid);
+    return user;
+  }
+
+  @Mutation((_returns) => [User], {
+    description: 'Creates Weekly Debriefs for all users',
+  })
+  async getWeeklyDebrief() {
+    const users = await WeeklyDebriefRepo.createWeeklyDebriefs();
```
alanna-zhou, line 69: why is this variable named users, i thought it was a weekly debrief? you can just `return await ...`

alanna-zhou (dismissed):
nice!!! keep up the good work~

src/repos/NotificationRepo.ts
```diff
@@ -101,10 +101,91 @@ const sendNewArticleAndroidNotification = async (
...
+  const apnProvider = new apn.Provider({
+    token: {
+      key: process.env.APNS_TOKEN_PATH,
+      keyId: process.env.APNS_KEY_ID,
+      teamId: process.env.APNS_TEAM_ID,
+    },
+  });
+
+  const notifTitle = 'Your Weekly Debrief is Ready';
+  const notifBody = 'Click to check out whats new in Volume this week!';
```
chalo2000, line 121: what's

src/repos/NotificationRepo.ts
```diff
@@ -101,10 +101,91 @@ const sendNewArticleAndroidNotification = async (
...
+
+/**
+ * Send Android push notification to {userID} that their weekly debrief is ready.
+ */
+const sendWeeklyDebriefAndroidNotification = async (user: User): Promise<void> => {
+  // Get device token of user
+  const { deviceToken } = user;
+
+  const notifTitle = 'Your Weekly Debrief is Ready';
+  const notifBody = 'Click to check out whats new in Volume this week!';
```
chalo2000, line 155: what's

chalo2000: these constants are the same as the one in the Android function, make sure to factor these out somewhere else (constants file?)

src/repos/NotificationRepo.ts
```diff
@@ -122,8 +203,25 @@ const notify = async (articleIDs: string[]): Promise<void> => {
...
 
+/**
+ * Send notifications for new articles have been posted by publications.
+ */
+const notifyWeeklyDebrief = async (users: User[]): Promise<void> => {
+  users.forEach(async (user) => {
+    if (user.deviceType === ANDROID) {
+      sendWeeklyDebriefAndroidNotification(user);
+    }
+    if (user.deviceType === IOS) {
```
chalo2000, line 214: else if, you can't ever have both of these cases so it doesn't make sense to check if for both

src/repos/UserRepo.ts
```diff
@@ -16,17 +18,18 @@ const createUser = async (
 
   // if no user, create a new one
```
chalo2000, line 19: can we have these comments start with capital letters?

src/repos/WeeklyDebriefRepo.ts
chalo2000: That's not my point, storing `readArticles` and `numReadArticles` is redundant, we just want to store `user.readArticles` since if we want to get what we're currently storing under `readArticles` and `numReadArticles` when can then just do `.slice(0,2)` and `.length`. If this is more intuitive to you @kidzegeye then feel free to keep it but I'd definitely be confused seeing that the user's `readArticles` and a weekly debrief's `readArticles` are not the same thing especially when this behavior is undocumented. If they were the same and then when I get a weekly debrief I do the splicing and the length operation and have it commented in the function doing it, then it is much clearer. Think about what we'll need to do in the future if we decide to have the weekly debrief UI show the three of the articles a user read over the course of the past week rather than two - now you're gonna have to manually update each of the objects and that will be a nightmare. This should be fixed before I approve this, this will lead to inflexibility in the future.

chalo2000 (changes requested):
The model structure needs to be changed before I can approve this - see my comments.

tedim52:
The PR with notifications and the notifyWeeklyDebrief should be a separate PR.

chalo2000 (dismissed):
spoke with Kidus, since Weekly Debriefs are a one-to-one it's okay if we have the data be inconsistent since it's not being stored so changing it later from 2 articles to 3 articles would be simple

tedim52 (dismissed):
This PR needs to be split up according to[ comment](https://github.com/cuappdev/volume-backend/pull/43#issuecomment-1060077598). Undo the merge commit, I can help if needed.

src/repos/WeeklyDebriefRepo.ts
kidzegeye: yes

src/resolvers/UserResolver.ts
kidzegeye: createWeeklyDebriefs() returns a UserModel

kidzegeye: changed to `return await ...`

publications.json
```diff
@@ -226,7 +226,7 @@
          "bio":"A student-run magazine that discusses business issues, trends, and debates among college students and alumni, CBR provides insight into recent events and passes along advice from student entrepreneurs and industry leaders and pioneers.",
          "bioShort":"Cornell Business Review is the premier business publication at Cornell University.",
          "rssName":"CBRnow - Cornell Business Review",
-         "rssURL":"http://www.thecornellbusinessreview.com/cbrnow?format=rss",
+         "rssURL":"https://www.cornellappdev.com/?format=rss",
```
chalo2000, line 229: why is Cornell Business Review's rssURL the cornell appdev site

chalo2000 (approved):
LGTM please merge

alanna-zhou (approved):
lgtm :)

publications.json
kidzegeye: The link used to be down last semester, but it seems to be up now. Will fix

dishanest (approved):
bro just merge already.

alanna-zhou (approved):
this has been too many cycles of review already -- next time it'd be great if the PRs were in a more smaller size so reviewers can digest it a little better, but otherwise keep up the great work!

PR #45, opened 2022-02-11
Add covid filter to all article query functions

kidzegeye:
## Overview

Added a Covid-19 filter to all article query functions in order to allow IOS to submit the app to the App Store.

## Changes Made

Added a filter on key words related to Covid-19 that checks the titles of queried articles. `getArticleByID`, `getArticleByIDs`, `getAllArticles`, `getArticlesByPublicationID`,`getArticlesByPublicationIDs`, `getArticlesAfterDate`, and `getTrendingArticles` have been filtered.

## Test Coverage

Manual testing was done by checking if Covid articles from before the change were hidden after the change was applied.

## Next Steps

Might need to add more key words to the filter in case that the ones I've added aren't broad enough.

### DIFF
--- a/src/repos/ArticleRepo.ts
+++ b/src/repos/ArticleRepo.ts
@@ -5,24 +5,35 @@ import { DEFAULT_LIMIT, MAX_NUM_DAYS_OF_TRENDING_ARTICLES } from '../common/cons
 import { PublicationModel } from '../entities/Publication';
 
 const getArticleByID = async (id: string): Promise<Article> => {
-  return ArticleModel.findById(new ObjectId(id));
+  const article = ArticleModel.findById(new ObjectId(id));
+  const covidCheck = isCovidArticle((await article).title);
+  if(!covidCheck){
+    return article;
+  }
+  return null;
 };
 
 const getArticlesByIDs = async (ids: string[]): Promise<Article[]> => {
   return Promise.all(ids.map((id) => ArticleModel.findById(new ObjectId(id)))).then((articles) => {
     // Filter out all null values that were returned by ObjectIds not associated
     // with articles in database
-    return articles.filter((article) => article !== null);
+    return articles.filter((article) => article !== null && !isCovidArticle(article.title));
   });
 };
 
 const getAllArticles = async (limit = DEFAULT_LIMIT): Promise<Article[]> => {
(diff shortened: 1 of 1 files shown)


src/repos/ArticleRepo.ts
```diff
@@ -103,6 +116,12 @@ const incrementShoutouts = async (id: string): Promise<Article> => {
   return article.save();
 };
 
+const isCovidArticle = (title: string): boolean => {
+  const filter = new Filter();
+  filter.addWords("covid", "covid-19", "coronavirus", "test", "mask", "masks", "pandemic", "tests", "omicron","endemic")
```
chalo2000, line 121: Does this filter automatically take care of lowercasing text? So we don't miss 'COVID' because the word we added is 'covid'.

Also, as discussed before, adding 'test' and 'tests' may be a little too restrictive as it could block articles about actual exams.

chalo2000 (approved):
I noticed that these worlds are only being checked in the title; I'll approve this for now because I don't want us to be blocked from the store but we should eventually do some checking in the body of the article (if possible through the RSS feeds) so that we can do a better job and catch articles easily without having to put restrictive words into the filter.

src/repos/ArticleRepo.ts
kidzegeye: We only have access to the article bodies in the microservice. I think it could be possible to do the check there and still not damage the article data.

tedim52, line 121: While the goal currently is to remove COVID articles, I don't think the code should have a notion that it's removing COVID articles - hopefully COVID will be gone soon and volume-backend will be around for longer. Generally, the task here is to remove "bad" articles and currently "bad" for us are articles with COVID related content. I suggest 2 changes: 1. Move all the "bad" title words to a list of words in the constants file. 2. Rename the method `isCovidArticle` to `isBadArticle` or something similar. This way when we no longer care to remove COVID content, rather than deprecate the function, you just remove the words in the constants file and you don't have to remove the function/task of removing "bad" articles. If one day Apple, comes along and says we have to remove all articles related to Microsoft - you don't change the function, you just add "Microsoft" to the list of constant words.

tedim52: @chalo2000 @dishanest feel free to comment on this

PR #51, opened 2022-09-15
Pagination Implementation

Archit404Error:
## Overview
Adds `offset` and `limit` params to the `getAllArticles`, `getArticlesByPublicationSlug`, `getArticlesByPublicationSlugs`, `getArticlesByPublicationID`, and `getArticlesByPublicationIDs` queries. 

The documentation of each corresponding function has been updated to show that limiting and offsets are supported. Default limit is 25 as specified in `constants.ts`. Default offset is 0, also specified in `constants.ts`.

There are also some invariants/constraints on the limit and offset values:
- Offset >= 0
- Limit > 0
- If offset + limit > articles in DB, the number of articles returned < limit (articles from between offsets `offset` to `max_offset` (inclusive) are returned)
- If offset > articles in DB, an empty list is returned

## Changes Made
- Utilized mongoose's `.skip()` to represent the offset functionality
- Also used `.limit()` for limit in repo functions that didn't have limit before

## Test Coverage
- Ran manual tests on all queries
- Tested `offset + limit < len(articles)1
- Tested `offset > len(articles)`
- Tested `offset + limit > len(articles)`
- Tested `offset = 0`, `limit = 0`
- Tested negative offsets and limits

All tests succeeded with intended behavior according to aforementioned constraints

- Implementing pagination on iOS/Android

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@
 *build/*
 nohup.out
 *src/certificates*
+Makefile
+start.sh
--- a/src/common/constants.ts
+++ b/src/common/constants.ts
@@ -1,7 +1,18 @@
 export const ANDROID = 'ANDROID';
 export const DAYS_IN_WEEK = 7;
 export const DEFAULT_LIMIT = 25.0;
-export const FILTERED_WORDS = ['covid-19', 'covid', 'coronavirus', 'masks', 'mask','pandemic','test','testing','tests'];
+export const DEFAULT_OFFSET = 1;
+export const FILTERED_WORDS = [
+  'covid-19',
+  'covid',
+  'coronavirus',
+  'masks',
+  'mask',
+  'pandemic',
+  'test',
+  'testing',
+  'tests',
+];
 export const IMAGE_ADDRESS = 'https://raw.githubusercontent.com/cuappdev/assets/master/volume';
 export const IOS = 'IOS';
 export const IS_FILTER_ACTIVE = true;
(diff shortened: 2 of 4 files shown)


src/common/constants.ts
```diff
@@ -1,7 +1,18 @@
 export const ANDROID = 'ANDROID';
 export const DAYS_IN_WEEK = 7;
 export const DEFAULT_LIMIT = 25.0;
-export const FILTERED_WORDS = ['covid-19', 'covid', 'coronavirus', 'masks', 'mask','pandemic','test','testing','tests'];
+export const DEFAULT_OFFSET = 1;
```
kidzegeye, line 4: Why is the default offset 1? Would that skip the first article each query?

kidzegeye (commented):
Hey! This looks great, just follow up on the comment I added and I'll approve

src/common/constants.ts
Archit404Error: Updated should be good to go now

kidzegeye (approved):
LGTM!

chalo2000 (approved):
LGTM but writing that your testing was manual is not enough. I need to be able to know what you did to test at the very least. If there were test cases, this would be obvious but we have already discussed we'll setup an automated Github Actions checker once we run out of features to finish up on the backend (in the next 2-3 weeks~). Write exactly what you manually did in testing coverage so we can reference it when writing the test cases; e.g. did you test negative offsets and sizes? What happens if I have 50 data points, a size of 5, and choose an offset for 11? Will that just give a 500 error or could it possibly cause server issues again?

.gitignore
```diff
@@ -5,3 +5,5 @@
 *build/*
 nohup.out
 *src/certificates*
+Makefile
```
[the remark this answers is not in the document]
Archit404Error: nothing, I have personal makefile scripts to ssh into dev & prod, j didn't want to commit those

PR #52, opened 2022-09-21
Implement notifications for Weekly Debrief

kidzegeye:
## Overview

Add notifications for weekly debriefs on a weekly cron job

## Changes Made

*Add Firebase notifications to weekly debrief creation
*Linted code and resolve deprecation with bodyParser()
*Added explicit null return to getArticleByID()
*Add distinction in notification calling between prod and dev
  -Dev initiates weekly debrief notifs on calls to getWeeklyDebrief()
## Test Coverage

1. Asked Han to build Volume's main branch on my phone to get my phone's device id
2. Created a user on my local branch containing my device id
3. Queried getWeeklyDebrief() to release the week's weekly debrief
4. Confirmed that my phone received a notification


src/app.ts
```diff
@@ -20,12 +21,13 @@ const main = async () => {
 
   await dbConnection();
 
-  var bodyParser = require('body-parser');
+  // eslint-disable-next-line vars-on-top, global-require, @typescript-eslint/no-var-requires
```
chalo2000, line 24: what is this linter disable doing?

src/app.ts
```diff
@@ -20,12 +21,13 @@ const main = async () => {
...
+  const bodyParser = require('body-parser');
 
-  var app = Express();
+  const app = Express();
 
-  app.use(bodyParser.urlencoded({ extended: false }))
-  app.use(bodyParser.json())
+  app.use(bodyParser.urlencoded({ extended: false }));
+  app.use(bodyParser.json());
```
chalo2000, line 30: should probably have the linter detect no semicolons too if that's possible

src/app.ts
```diff
@@ -38,27 +40,34 @@ const main = async () => {
   });
 
   // Setup Firebase Admin
-  if (process.env.NODE_ENV === 'production') {
-    admin.initializeApp({
-      credential: admin.credential.cert(process.env.FCM_AUTH_KEY_PATH),
-    });
-  }
+  admin.initializeApp({
```
chalo2000, line 43: why is this no longer covered by a production environment variable?

src/repos/ArticleRepo.ts
```diff
@@ -27,6 +27,8 @@ const getArticleByID = async (id: string): Promise<Article> => {
     if (!isArticleFiltered(article)) {
       return article;
     }
+
+    return null;
```
chalo2000, line 31: what would be returned before?

src/repos/NotificationRepo.ts
```diff
@@ -60,7 +60,52 @@ const notify = async (articleIDs: string[]): Promise<void> => {
   });
 };
 
+const sendWeeklyDebriefNotification = async (user: User): Promise<void> => {
+  // Get device token of user
+  const { deviceToken } = user;
+
+  const notifTitle = 'Your Weekly Debrief is Ready';
```
chalo2000, line 67: do you think there would be value in putting these const titles, bodies, messages, and options into a separate file to make it easy to find them when we need to change them if ever? do we plan to have any other kinds of notifications?

chalo2000 (changes requested):
Can you please include the PR what exactly you did to test under testing coverage? So if someone reads this PR and wants to do the testing steps you can, it is easy to replicate that it works. Also, lots of comments I left (aside from moving constants into another file if you think that is good) weren't mentioned in the PR so I am confused how it relates to implementing notifications. You'd usually want to put these under 'Changes Made'.

src/app.ts
kidzegeye: Will remove the disable and also bodyParser since it is deprecated now

src/app.ts
kidzegeye: added

src/app.ts
kidzegeye: I wasn't sure why it should be in the first place, as it makes testing notifications in dev not possible

src/repos/ArticleRepo.ts
kidzegeye: It would return null

src/repos/NotificationRepo.ts
kidzegeye: As of now the only notifications with constant titles is the weekly debrief. I believe magazines might have notifications (maybe not since they are semesterly), but they wouldn't have constant titles however.

src/app.ts
kidzegeye: Readded

Archit404Error (commented):
This looks good to me, just want to see a distinction between dev/prod/local environments such that:
- Prod: only send notifs based on cronjob scheduling
- Dev & local: send notifs to device upon mutation call

Archit404Error (approved):
Good to go!

src/repos/WeeklyDebriefRepo.ts
```diff
@@ -47,6 +48,11 @@ const createWeeklyDebriefs = async (): Promise<User[]> => {
   const userList = UserModel.find({}).then((users) =>
     Promise.all(users.map((user) => createWeeklyDebrief(user, creationDate, expDate))),
   );
+  if (process.env.NODE_ENV === 'dev') {
```
chalo2000, line 51: why should this only happen on dev? add a comment

chalo2000 (approved):
LGTM

src/repos/WeeklyDebriefRepo.ts
kidzegeye: Since on prod notifications will only be sent on the weekly cron job, to test out notifications on dev I added the notification handler call in the createWeeklyDebriefs mutation. This shouldn't be used on prod since this adds an await in an asynchronous function.

PR #53, opened 2022-09-25
Implement chronological sorting

Archit404Error:
## Overview

This is a follow up from #51, because we also want to sort our results chronologically prior to sending them over (we should be returning the 25 most recent articles when offset is 0 and limit is 25).

This PR implements that sorting, and also cleans up previous code used in the relevant methods that was pretty unclean/hard to read.

Relevant methods (all methods were within `ArticleRepo.ts`):
- `getAllArticles`
- `getArticlesByPublicationSlug`
- `getArticlesByPublicationSlugs`
- `getArticlesByPublicationID`
- `getArticlesByPublicationIDs`

## Changes Made

Added a simple `.sort()` call before the limit and offset is applied, in order to sort by date prior to returning results. Additionally, cleaned up `getArticlesByPublicationSlugs` so it satisfies the intended offset and limit behavior (before it was applying the limit/offset to every single publication in the slug array's articles, so it was actually returning `n * limit` results, where `n = len(slugs)`).

**NOTE: Made some changes to the ID methods but these are deprecated and will be removed soon. They work but for all intents and purposes are dead code while frontend moves to using slugs.**

## Test Coverage

Went through each of the resolvers corresponding to the methods listed above (the resolvers have the same name), and testing the limit/offset to make sure articles were sorted as expected. Additionally, tested the edge cases described in #51 once again to make sure that nothing has broken as a result of my changes.

### DIFF
--- a/src/repos/ArticleRepo.ts
+++ b/src/repos/ArticleRepo.ts
@@ -43,11 +43,35 @@ const getAllArticles = async (
   limit = DEFAULT_LIMIT,
 ): Promise<Article[]> => {
   return ArticleModel.find({})
+    .sort({ date: 'desc' })
     .skip(offset)
     .limit(limit)
     .then((articles) => {
       return articles.filter((article) => !isArticleFiltered(article));
     });
+};
+
+const getArticlesByPublicationSlug = async (
+  slug: string,
+  limit: number = DEFAULT_LIMIT,
+  offset: number = DEFAULT_OFFSET,
+): Promise<Article[]> => {
+  return ArticleModel.find({ 'publication.slug': slug })
+    .sort({ date: 'desc' })
+    .skip(offset)
+    .limit(limit);
+};
+
+const getArticlesByPublicationSlugs = async (
+  slugs: string[],
+  limit: number = DEFAULT_LIMIT,
+  offset: number = DEFAULT_OFFSET,
+): Promise<Article[]> => {
+  const uniqueSlugs = [...new Set(slugs)];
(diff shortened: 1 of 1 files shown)


kidzegeye (approved):
LGTM

src/repos/ArticleRepo.ts
```diff
@@ -43,61 +43,65 @@ const getAllArticles = async (
...
   offset: number = DEFAULT_OFFSET,
 ): Promise<Article[]> => {
-  const publication = await (await PublicationModel.findById(publicationID)).execPopulate();
-  return ArticleModel.find({ 'publication.slug': publication.slug })
+  return ArticleModel.find({ 'publication.slug': slug })
+    .sort({ date: 'desc' })
     .skip(offset)
-    .limit(limit)
-    .then((articles) => {
-      return articles.filter((article) => !isArticleFiltered(article));
```
chalo2000, line 63: what happened to returning the filtered articles?

chalo2000: I see we still have this for the publicationID - should we not also have a filter on the slugs? To prevent being blocked on the App Store?

chalo2000 (approved):
LGTM - I understand the need for the deprecated functions until iOS fully moves over to prevent breaking it but please take another look at getArticlesByPublicationSlug since I noticed this is no longer doing a filter. I may be mistaken if this is done somewhere else but let's ensure we don't get blocked on the App Store again. A test for getting articles that should be filtered would be good.

PR #54, opened 2022-09-29
Publication Slug Migration

Archit404Error:
## Overview
This PR migrates the `followPublication` and `unfollowPublication` to slugs rather than ids and thus fully deprecates `PublicationId`s in the backend codebase. Our reasoning for this decision was as follows:
- **Background:** Initially, all publications were uniquely identified by the id hash that MongoDB automatically generates for any new document
- **Situation:** It was really hard to debug across local, dev, and prod servers using this system, seeing as publishers had different ids on **every single DB**. We needed to figure out a better way to identify publishers uniquely across all databases
- **Decision:** Slugs (short unique strings to identify each publisher) had already been implemented in prior semesters, and we had partially transitioned to them. We decided to fully transition all publication references to slugs in order to phase out IDs once and for all and resolve issues caused by mixing IDs and slugs

## Changes Made
Removed the `PublicationId` class and replaced it with `PublicationSlug`, seeing as the class wasn't being used anywhere besides these mutations. Also wrote `followingSlugs.js` migration to be used by migrate-mongo to move existing users which have publication ids stored in their `followedPublications` field to storing publication slugs instead.

## Test Coverage

Utilized GraphQL playground to create a user, add some default publications using slugs, and then also test out the `followPublication` and `unfollowPublication` mutations. See screenshots of tests below:

Screen Shot 2022-09-29 at 10 57 30 AM
Screen Shot 2022-09-29 at 10 57 35 AM
Screen Shot 2022-09-29 at 11 03 35 AM
Screen Shot 2022-09-29 at 11 04 09 AM

Additionally, tested the migration by running it on the local database and ensuring that all following publication ids were converted to slugs and vice versa. Ran the commands via `migrate-mongo up followingSlugs` and `migrate-mongo down followingSlugs`. Upon migrating up and down, I used `mongosh` to interact with the local mongodb server and ensure documents had been updated as expected.

## Next Steps

If necessary, we might move from just storing publication slugs to an entire subdocument within `followedPublications` if this provides a speedup to frontend. As of now based on my current understanding this wouldn't actually make a difference so will make another quick PR for that if it's necessary.

### DIFF
--- a/migrations/followingSlugs.js
+++ b/migrations/followingSlugs.js
@@ -1,0 +1,31 @@
+module.exports = {
+    async up(db) {
+        /**
+         * Convering user followed publication ids to slugs
+         */
+        const allUsers = await db.collection('users').find({}).toArray();
+        allUsers.forEach(async (user) => {
+            const pubIds = user.followedPublications;
+            await db.collection('users').updateOne({ _id: user._id }, { $set: { followedPublications: [] } });
+            pubIds.forEach(async (id) => {
+                const pubData = await db.collection('publications').findOne({ _id: id })
+                await db.collection('users').updateOne({ _id: user._id }, { $push: { followedPublications: pubData.slug } });
+            });
+        })
+    },
+
+    async down(db) {
+        /**
+         * Convering user followed publication slugs to ids
+         */
+        const allUsers = await db.collection('users').find({}).toArray();
+        allUsers.forEach(async (user) => {
+            const pubSlugs = user.followedPublications;
+            await db.collection('users').updateOne({ _id: user._id }, { $set: { followedPublications: [] } });
+            pubSlugs.forEach(async (slug) => {
+                const pubData = await db.collection('publications').findOne({ slug: slug })
+                await db.collection('users').updateOne({ _id: user._id }, { $push: { followedPublications: pubData._id } });
+            });
+        })
+    }
+}
--- a/migrations/followingSlugs.js
+++ b/migrations/followingSlugs.js
@@ -1,7 +1,7 @@
 module.exports = {
     async up(db) {
         /**
-         * Convering user followed publication ids to slugs
+         * Converting user followed publication ids to slugs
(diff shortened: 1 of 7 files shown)


migrations/followingSlugs.js
```diff
@@ -0,0 +1,31 @@
+module.exports = {
+    async up(db) {
+        /**
+         * Convering user followed publication ids to slugs
```
chalo2000, line 4: Converting

src/repos/UserRepo.ts
```diff
@@ -39,10 +36,11 @@ const createUser = async (
 /**
  * Adds publication of pubID to user represented by uuid's followedPublications.
```
chalo2000, line 37: change the docstrings

chalo2000 (dismissed):
LGTM; you shouldn't link the Notion since we want to make sure that changes in our internal tools don't affect the history of the codebase (feel free to just copy paste the reasoning for the decision from Notion to here). Also some minor comment fixes.

migrations/followingSlugs.js
Archit404Error: lol

src/repos/UserRepo.ts
Archit404Error: word

chalo2000, line 37: missed some

src/repos/UserRepo.ts
```diff
@@ -51,16 +49,14 @@ const followPublication = async (uuid: string, pubID: string): Promise<User> =>
 /**
  * Deletes publication of pubID from user represented by uuid's followedPublications.
```
chalo2000, line 50: another

chalo2000 (dismissed):
LGTM

chalo2000 (approved):

kidzegeye (approved):
LGTM

PR #56, opened 2022-10-12
Make filtering consistent with all article functions

kidzegeye:
## Overview

Add Article filter to all Article repo functions that return articles

## Changes Made

- Added article filtering to getArticlesByPublicationSlug and getArticlesByPublicationSlugs
- Turned off article filtering
- Removed getArticlesByPublicationID and getArticlesByPublicationIDs to be consistent with https://github.com/cuappdev/volume-backend/pull/54
- Made IS_FILTER_ACTIVE an environment variable

## Test Coverage

Tested getArticlesByPublicationSlug and getArticlesByPublicationSlugs with articles that were flagged as profane and were not flagged as profane to see if the articles would be filtered correctly


src/common/constants.ts
```diff
@@ -15,5 +15,5 @@ export const FILTERED_WORDS = [
 ];
 export const IMAGE_ADDRESS = 'https://raw.githubusercontent.com/cuappdev/assets/master/volume';
 export const IOS = 'IOS';
-export const IS_FILTER_ACTIVE = true;
+export const IS_FILTER_ACTIVE = false;
```
chalo2000, line 18: ironic that the branch is called filter articles but filtering is off

precisely why I said to make this an environment variable in your previous commit - consider making an issue to do this in the future in a small PR

chalo2000 (approved):
LGTM - change the PR title

src/repos/ArticleRepo.ts
```diff
@@ -96,21 +97,6 @@ const getArticlesByPublicationID = async (
...
-  const uniquePubIDs = [...new Set(publicationIDs)].map((id) => new ObjectId(id));
-  console.log(uniquePubIDs);
-  const pubSlugs = await PublicationModel.find({ _id: { $in: uniquePubIDs } }).select('slug');
-  return getArticlesByPublicationSlugs(
-    pubSlugs.map((pub) => pub.slug),
-    limit,
-    offset,
-  );
-};
-
```
Archit404Error, line 113: Are we 100% sure the existing iOS app does not make use of this?

Archit404Error (commented):
This looks good but not sure why we're deleting `getArticlesByPublicationIDs` in this PR, see comment below

src/repos/ArticleRepo.ts
```diff
@@ -96,21 +97,6 @@ const getArticlesByPublicationID = async (
...
-): Promise<Article[]> => {
-  const uniquePubIDs = [...new Set(publicationIDs)].map((id) => new ObjectId(id));
-  console.log(uniquePubIDs);
-  const pubSlugs = await PublicationModel.find({ _id: { $in: uniquePubIDs } }).select('slug');
-  return getArticlesByPublicationSlugs(
-    pubSlugs.map((pub) => pub.slug),
-    limit,
-    offset,
-  );
-};
```
Archit404Error, line 112: Are we 100% sure the existing iOS app does not make use of this?

kidzegeye: Will readd, I thought it was removed in main

.envrctemplate
```diff
@@ -1,3 +1,4 @@
 export NODE_ENV=[node env]
 export APP_PORT=[port]
-export MONGO_ADDRESS=[db uri]
\ No newline at end of file
+export MONGO_ADDRESS=[db uri]
+export IS_FILTER_ACTIVE = false
```
chalo2000, line 4: the template should not hold the actual value, I would replace these with
```
export MONGO_ADDRESS=[mongo address]
export IS_FILTER_ACTIVE=[is filter active]
```

chalo2000: also include that you added environment variables in the PR description

chalo2000: I also see MONGO_ADDRESS isn't used anywhere in the code for this PR?

chalo2000 (approved):
review my comments and merge if nothing pressing

Archit404Error (approved):
Agree with Gonzalo's comments, good to go after that

PR #61, opened 2022-11-02
Add logos as a magazine publisher

Archit404Error:
## Overview

A simple PR that adds logos as a magazine publisher.

## Changes Made

Added logos' information to `publications.json`

## Test Coverage

Ran the micro service & backend locally and ensured logos was added to DB (see screenshot below)

## Screenshots 
Screen Shot 2022-11-01 at 9 59 50 PM

### DIFF
--- a/publications.json
+++ b/publications.json
@@ -224,7 +224,27 @@
             }
          ],
          "websiteURL":"https://threadcornell.com/"
-    },
+      },
+      {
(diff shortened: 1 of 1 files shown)


tedim52:
LETSSSS GOOO.

kidzegeye (approved):
Ideally we would deprecate this json since the publishers should already be populated in the database via the microservice. That's a later-problem though

chalo2000 (approved):
LGTM and agree with Kidus' point; one day this data will be collected from some form that publications fill out and we serve directly through the microservice.

PR #62, opened 2022-11-03
Add differentiation between iOS and Android notifs

Archit404Error:
## Overview

This PR creates separate notif types for both android and iOS based on a user's device type, as requested by the android subteam (because android needs a different data format than what we're sending to iOS).

It also factors out all duplicated code across various notification functions and centralizes them in a single `sendNotif` function to increase maintainability in the future and convert all notification types to the new format for android notifs.

## Changes Made
- Use user's device type in order to determine what type of notif to send
- Factor out notif code to prevent code duplication using JS spread syntax (the ... operator)

## Test Coverage
I first modified just the article notif function and ensured that the objects were constructed properly by testing with dummy values, since that's the only thing that was changed. I then factored out the code and ensured that the data being sent was the same as before.

Also asked Kidus & Ben to pull & try sending notifs on iOS/android, which worked

### DIFF
--- a/src/repos/NotificationRepo.ts
+++ b/src/repos/NotificationRepo.ts
@@ -17,18 +17,33 @@ const sendNewArticleNotification = async (
   const notifTitle = publication.name;
   const notifBody = article.title;
 
-  const message = {
-    notification: {
-      title: notifTitle,
-      body: notifBody,
-    },
-    data: {
-      userID: user.uuid,
-      articleID: article.id,
-      articleURL: article.articleURL,
-      notificationType: 'new_article',
-    },
+  const notifData = {
+    title: notifTitle,
+    body: notifBody,
   };
+
+  const metaData = {
+    userID: user.uuid,
+    articleID: article.id,
+    articleURL: article.articleURL,
+    notificationType: 'new_article',
+  };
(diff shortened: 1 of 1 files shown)


kidzegeye (approved):
LGTM, might want to add the changes for magazine notifications though, or we can make one function for building all content-types' notification message

kidzegeye (approved):
LGTM

src/repos/NotificationRepo.ts
```diff
@@ -8,28 +8,51 @@ import ArticleRepo from './ArticleRepo';
...
+
+  // Determine notif format based on device type
+  if (user.deviceType === 'IOS') {
+    message = {
+      notification: notifData,
+      data: metaData,
+    };
+  } else {
+    message = {
+      data: {
```
chalo2000, line 49: Android should migrate to iOS's format later so that we can have consistent responses. @ckdesir

chalo2000 (approved):
nice notif refactor

src/repos/NotificationRepo.ts
ckdesir: Don't think that will be possible.

PR #63, opened 2022-11-09
Add Jest testing

kidzegeye:
## Overview

Add Jest testing capabilities

## Changes Made

* Created ArticleFactory and PublicationFactory in order to create fake Articles and query Publications for testing
* Added FactoryUtils from Resell to have utility functions for Factory classes
* Added `ArticleRepo` test cases for `getAllArticles`, `getArticle(s)ByID(s)`, `getArticlesBySlug(s)`, `getArticleAfterDate`, `searchArticle`, and `incrementShoutouts`
* Added `PublicationRepo` test cases for `getAllPublications`, `getMostRecentArticle`, `getNumArticles`, `getPublicationBySlug`, and `getShoutouts`
* Added a test database connection to volume-test database
* Added pre-commit hook to `npm run jest`

## Test Coverage

Ran jest test cases, ran to completion

Add test cases for MagazineRepo functions

<details>

  <summary>Jest test script output</summary>

image

</details>


tedim52:
love these tests

src/tests/data/PublicationFactory.ts
```diff
@@ -0,0 +1,13 @@
+/* eslint-disable no-underscore-dangle */
+import { ObjectId } from 'mongodb';
+import { PublicationModel } from '../../entities/Publication';
+
+class PublicationFactory {
+  public static async getRandomPublication() {
+    const pubs = await PublicationModel.aggregate().sample(1);
+    const pub = await PublicationModel.findById(new ObjectId(pubs[0]._id));
```
Archit404Error, line 8: Wouldn't `PublicationModel.aggregate().sample(1);` return the object as the first element in the array anyway? Why are we taking the ID and then looking up the object again?

src/tests/data/FactoryUtils.ts
```diff
@@ -0,0 +1,64 @@
...
+      case TimeUnits.Days:
+        return amount * 1000 * 60 * 60 * 24;
+      case TimeUnits.Months:
+        return amount * 1000 * 60 * 60 * 24 * 30;
+      case TimeUnits.Years:
+        return amount * 1000 * 60 * 60 * 24 * 365;
+      default:
+        throw new Error('Invalid TimeUnit enum value');
+    }
+  }
```
Archit404Error, line 63: What's the point of this method? I don't see it being used anywhere?

src/tests/data/ArticleFactory.ts
```diff
@@ -0,0 +1,86 @@
...
+    fakeArticle.imageURL =
+      'https://raw.githubusercontent.com/cuappdev/assets/master/volume/placeholders//nooz.png';
+    fakeArticle.isFiltered = false;
+    fakeArticle.publication = fakePublication;
+    fakeArticle.publicationSlug = 'nooz';
+    fakeArticle.title = 'Op-Ed: The Real World Has Stingrays. So Why Doesn’t the Swim Test?';
+    fakeArticle.isTrending = false;
+    fakeArticle.nsfw = false;
+    fakeArticle.shoutouts = 0;
+    fakeArticle.trendiness = 0;
```
Archit404Error, line 57: This should eventually be refactored so we can initialize the object with these params instead of having to write verbose object assignment like this

src/tests/article.test.ts
```diff
@@ -0,0 +1,137 @@
...
+import ArticleFactory from './data/ArticleFactory';
+import PublicationFactory from './data/PublicationFactory';
+
+import { dbConnection, disconnectDB } from './data/TestingDBConnection';
+
+// Maps an array of mongo documents [x] to an array of x.[val]
+function mapToValue(arr, val) {
+  return arr.map((x) => {
+    return x[val];
+  });
```
Archit404Error, line 16:
```suggestion
  return arr.map((x) => x[val]);
```

src/tests/article.test.ts
```diff
@@ -0,0 +1,137 @@
...
+});
+
+beforeEach(async () => {
+  await ArticleModel.deleteMany({});
+});
+
+afterAll(async () => {
+  await ArticleModel.deleteMany({}).then(() => {
+    disconnectDB();
+  });
```
Archit404Error, line 46: I think we can do this here because rn we're just passing in a func which calls `disconnectDB` which is the same as just passing disconnectDB in directly:

```suggestion
  await ArticleModel.deleteMany({}).then(disconnectDB);
```

Archit404Error (changes requested):
This looks super solid & it's awesome that we'll have solid regression testing now! I left a few nits/questions, also were you planning on adding the pre-commit hook + readme changes in a different PR? I don't see those here

src/tests/data/ArticleFactory.ts
kidzegeye: I think most test cases will use the articles created by the fake() function. I removed the template for now

kidzegeye:
I can see the precommit hook change in the .github folder

Archit404Error (approved):
This looks good to go! It looks like the hook changes maybe weren't pushed to remote before: Screen Shot 2022-11-09 at 11 28 18 PM

Regardless, good work on this, it looks great!

src/tests/article.test.ts
```diff
@@ -0,0 +1,133 @@
...
+import { ArticleModel } from '../entities/Article';
+import ArticleRepo from '../repos/ArticleRepo';
+import PublicationRepo from '../repos/PublicationRepo';
+import ArticleFactory from './data/ArticleFactory';
+import PublicationFactory from './data/PublicationFactory';
+
+import { dbConnection, disconnectDB } from './data/TestingDBConnection';
+
+// Maps an array of mongo documents [x] to an array of x.[val]
+function mapToValue(arr, val) {
```
snajima, line 13: this feels like something that should be in `FactoryUtils.ts`

snajima (dismissed):
I'll give the green flag since @Archit404Error did, but `mapToValue` and `setAllToValue` feels like functions you would use in your other test files too

Archit404Error (changes requested):
Good work on this, nice that you added in the publication tests too but in the future I'd recommend splitting into 2 separate PRs! Left a few comments here and there, once those are addressed and/or fixed should be good to go!

src/tests/data/FactoryUtils.ts
```diff
@@ -0,0 +1,13 @@
...
+    return Array(n).fill(null).map(fn);
+  }
+
+  public static mapToValue(arr, val) {
+    return arr.map((x) => x[val]);
+  }
+  public static compareByDate(a, b) {
+    return -1 * (new Date(a.date).getTime() - new Date(b.date).getTime());
+  }
+}
```
Archit404Error, line 12: Usually a good idea to document helper methods like these since they'll be used all throughout our tests so I'd suggest docstrings for each of these

Archit404Error (approved):
Good to go! Just left one small documentation request but otherwise I think everything look great!

snajima (approved):
lgtm!

src/tests/data/ArticleFactory.ts
```diff
@@ -0,0 +1,65 @@
...
+    const examplePub = await PublicationFactory.getRandomPublication();
+
+    fakeArticle.articleURL = faker.internet.url();
+    fakeArticle.date = faker.date.past();
+    fakeArticle.imageURL = faker.image.cats();
+    fakeArticle.isFiltered = false;
+    fakeArticle.publication = examplePub;
+    fakeArticle.publicationSlug = examplePub.slug;
+    fakeArticle.title = faker.commerce.productDescription();
+    fakeArticle.isTrending = _.sample([true, false]);
```
chalo2000, line 57: wondering why the choice of this underscore library

src/tests/data/FactoryUtils.ts
```diff
@@ -0,0 +1,13 @@
+class FactoryUtils {
+  public static create<T>(n: number, fn: () => T): T[] {
+    return Array(n).fill(null).map(fn);
+  }
+
+  public static mapToValue(arr, val) {
+    return arr.map((x) => x[val]);
+  }
+  public static compareByDate(a, b) {
+    return -1 * (new Date(a.date).getTime() - new Date(b.date).getTime());
```
chalo2000, line 10: why not just flip the subtraction? is there a reason why it's being done like this?

chalo2000 (approved):
LGTM - just try to organize your tests so they're separated in the Arrange-Act-Assert format so it is very clear where you are actually running what you're testing and how you're testing it.

.github/hooks/pre-commit
```diff
@@ -0,0 +1 @@
+npm run jest
```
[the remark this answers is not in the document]
kidzegeye: yeah

src/tests/article.test.ts
```diff
@@ -0,0 +1,168 @@
...
+});
+
+afterAll(async () => {
+  ArticleModel.deleteMany({}).then(disconnectDB);
+});
+
+describe('getAllArticle tests:', () => {
+  test('getAllArticles - No articles', async () => {
+    const getArticlesResponse = await ArticleRepo.getAllArticles();
+    expect(getArticlesResponse).toHaveLength(0);
```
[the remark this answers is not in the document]
kidzegeye: will look into this pattern

src/tests/data/ArticleFactory.ts
kidzegeye: it has some nice out of the box functions for arranging test data

src/tests/data/FactoryUtils.ts
kidzegeye: no reason, i think i realized it was sorted the other way and added a -1

PR #66, opened 2023-03-12
Add magazines to weekly debrief

isaachan100:
## Overview
Added magazines to weekly debrief feature
Add jest testing for magazines and users

## Changes Made
- updated weekly debrief and user model types
- changed logic in controllers to support magazines in weekly debrief
- created mutation: readMagazine(...): User
- updated mutation: getWeeklyDebrief(...): [WeeklyDebrief!]!

Miscellaneous
- added rules to eslintrc to suppress `import/named` and `import/prefer-default-export` warnings
- updated gitignore to include secrets folder

Jest testing:
- added `MagazineRepo` test cases for `getAllMagazines`, `getMagazine(s)ByID(s)`, `getMagazinesByPublicationSlug(s)`, `getMagazinesBySemester`, `getFeaturedMagazines`, and `incrementShoutouts`
- added `UserRepo` test cases for `getUserByUUID`, `(un)followPublication`, `incrementShoutouts`, `incrementBookmarks`, `appendReadMagazine`, and `appendReadArticle`.
- tested Weekly Debrief indirectly through UserRepo test cases

## Test Coverage

ran the two new mutations locally and verified the outputs
all jest test cases passed

PR #68, opened 2023-03-14
Implement reshuffle articles

SashaLoayza:
## Overview

Implemented reshuffling of articles, method called 'getShuffledArticlesByPublicationSlugs' in ArticleRepos and ArticleResolvers.

## Changes Made

Return a list of Article objects sorted by date, but each publisher has to show up at least once before articles by the same publisher shows up again
e.g. :
publisher 1 has articles 1.1 (published jan 1), 1.2 (published jan 6)
publisher 2 has articles 2.1 (published jan 5), 2.2 (published jan 4)
publisher 3 has articles 3.1 (published jan 2), 3.2 (published jan 3)
getShuffledArticlesByPublicationSlugs should return [1.1, 3.1, 2.1, 3.2, 2.2, 1.2]

The time range that articles should be retrieved from can be specified with timeRange (integer representing months since current date).

The query takes required parameter of a list of publication slugs (strings) called 'slugs', and optional parameters of 'limit', 'offset', and 'timeRange'.

## Test Coverage

Manually tested locally on GraphQL playground. Tested with two and three different publication slugs.

## Next Steps 

May need to add resolvers for querying by list of publication IDs, which will be a simple map to this function.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -9,7 +9,8 @@ module.exports = {
     '@typescript-eslint/explicit-module-boundary-types': 0,
     '@typescript-eslint/no-unused-vars': 0,
     'semi-style': ['error', 'last'],
-    'import/prefer-default-export': 'off',
+    'import/named': 0,
+    'import/prefer-default-export': 0,
   },
   // This gets rid of weird react error for non-react project
   settings: {
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -9,6 +9,7 @@ module.exports = {
     '@typescript-eslint/explicit-module-boundary-types': 0,
     '@typescript-eslint/no-unused-vars': 0,
     'semi-style': ['error', 'last'],
+    'import/prefer-default-export': 'off',
   },
   // This gets rid of weird react error for non-react project
   settings: {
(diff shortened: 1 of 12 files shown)


snajima (approved):
lgtm, let's just address the same issues as #66, then get this merged!

snajima:
for the sake of development we're skipping jest cases for this endpoint – also the tests would more so assess the algorithm rather than the backend/db queries functioning (since that's covered in the general get articles test cases). good point tho ! @isaachan100

snajima (changes requested):
for the sake of development we're skipping test cases for this endpoint – also since the tests would more so assess the shuffling algorithm rather than the backend/db queries functioning (bc that's already covered in the get all articles tests). good point tho! @isaachan100 

i'm gonna need you to revert your changes in `.eslintrc.js`, `package-lock.json`, and `package.json` if these were addressed by isaac though :)

isaachan100 (approved):
looks good!

isaachan100 (commented):
lgtm!

snajima (approved):
let's go sasha!!!

isaachan100 (commented):
lgtm!

isaachan100 (approved):
looks good!

PR #70, opened 2023-03-23
Add 3 new publications to publications.json

SashaLoayza:
Add Collective X, Cornell Healthcare Review, and Cornell Daily Sun to publications.json
## Changes Made
Collective X has no RSS feed (set to null as per the existing pattern) while Cornell Daily Sun and The Cornell HealthCare Review

## Related PRs or Issues (delete if not applicable)

Same PR in volume-backend.

### DIFF
--- a/publications.json
+++ b/publications.json
@@ -1,361 +1,440 @@
 {
-   "publications":[
-      {
-         "bio":"The Advocate promotes the dissemination of literature that examines youth-related issues with the goal of encouraging youth advocacy in local and federal politics.",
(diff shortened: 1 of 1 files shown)


snajima (approved):

SashaLoayza (commented):
Technically: I should fix the Healthcare Review URL, it is incorrect, and Daily Sun RSS feed url is null on micro service repo. 

However: The only parts of this file that the backend uses are the socialURLs, so it is not urgent to fix these.

isaachan100 (approved):

PR #71, opened 2023-03-24
Convert result of shuffling resolver to ArticleModel

SashaLoayza:
## Overview

I removed shadowing variables and fixed the functionality of the getShuffledArticlesByPublicationSlugs

MongoDB's aggregate functionality returns a json object, which does not contain the proper ID field (it contains _id without the getter needed to resolve from a GraphQL query). This caused queries to fail. 

## Changes Made

To fix the null id results in GraphQL queries, I converted each article JSON back into an ArticleModel object using the ArticleModel() constructor.

## Test Coverage

I tested that queries containing the article's ID and the nested publication's ID returned the expected result. I used console.log to view the articles before they were returned.

It may be worth it to refactor the getShuffledArticlesByPublicationSlugs function in ArticleRepos to use multiple .find queries rather than .aggregate to avoid the conversion to JSON and back to ArticleModel. This will also match the pattern of other functions and have less potential for unexpected behavior. However, this seems to work for now.

## Related PRs or Issues (delete if not applicable)
#68

### DIFF
--- a/src/repos/ArticleRepo.ts
+++ b/src/repos/ArticleRepo.ts
@@ -130,14 +130,14 @@ const getShuffledArticlesByPublicationSlugs = async (
       },
     },
   ]).then((articles) => {
-    const articlesByPub = {};
-    for (const article of articles) {
-      articlesByPub[article._id] = article.articles.filter(
+    const filteredArticles = {};
+    for (const a of articles) {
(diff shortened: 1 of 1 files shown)


isaachan100 (approved):
lgtm!

snajima (approved):
🤩 🎉

PR #74, opened 2023-04-11
Merge main into release

isaachan100:
## Overview

Merge dev server updates into prod server

## Test Coverage

- iOS verified on dev server


SashaLoayza (approved):

src/repos/ArticleRepo.ts
```diff
@@ -83,6 +84,81 @@ const getArticlesByPublicationSlugs = async (
...
+  limit: number = DEFAULT_LIMIT,
+  offset: number = DEFAULT_OFFSET,
+  timeRange = 12, // int value in months before current date to get articles from
+): Promise<Article[]> => {
+  // return a list of ArticleModels sorted by date, but each publisher has to show up at least once
+  // before articles by the same publisher shows up again
+  // e.g. publisher 1 has articles 1.1 (published jan 1), 1.2 (published jan 6)
+  //      publisher 2 has articles 2.1 (published jan 5), 2.2 (published jan 4)
+  //      publisher 3 has articles 3.1 (published jan 2), 3.2 (published jan 3)
+  // the function should return [1.1, 3.1, 2.1, 3.2, 2.2, 1.2]
```
Archit404Error, line 98: This should ideally be a docstring, anybody calling this function should be able to see this information, so it's better to structure this as a docstring with an `@returns` than as multiple single-line comments

src/repos/ArticleRepo.ts
```diff
@@ -83,6 +84,81 @@ const getArticlesByPublicationSlugs = async (
...
+  ]).then((articles) => {
+    const filteredArticles = {};
+    for (const a of articles) {
+      filteredArticles[a._id] = a.articles.filter(
+        (article) => article !== null && !isArticleFiltered(article),
+      );
+      mostByOnePub = Math.max(mostByOnePub, filteredArticles[a._id].length);
+    }
+    return filteredArticles;
+  });
```
Archit404Error, line 141: You should never be using `await` and `.then` together in an async function -- always stick to one or the other because they're different methods of resolving promises. Since you already `await` the results of the aggregation, you can just perform the for loop operation directly afterwards. Here's a rough outline of what this looks like:

```suggestion
  const articles = await ArticleModel.aggregate([
    {
      $match: {
        'publication.slug': { $in: uniqueSlugs },
        date: { $gte: since },
      },
    },
    {
      $sort: { date: -1 },
    },
    {
      $skip: offset,
    },
    {
      $limit: limit,
    },
    {
      $group: {
        _id: '$publication.slug',
        articles: { $push: '$$ROOT' },
      },
    },
  ])
  
    const articlesByPub = {};
    for (const a of articles) {
      filteredArticles[a._id] = a.articles.filter(
        (article) => article !== null && !isArticleFiltered(article),
      );
      mostByOnePub = Math.max(mostByOnePub, filteredArticles[a._id].length);
    }
```

Archit404Error (approved):
Approving because each of these these PRs have been individually approved. Some comments:
1. This is ~8 PRs being merged into release, in the future it'd be better to merge more incrementally (every 1-2 PRs) so that it's more clear what changes are being pushed to prod
2. I just skimmed over your code but I found 2 major issues with the way you're handling promise resolution in one part of your code and also a function docstring -- definitely make sure to resolve these issues via a PR on the dev server


Other than that, if you've tested everything on your end and made sure it works, you should be good to go!

PR #93, opened 2023-06-01
Implement cboard start and end dates

isaachan100:
## Overview

- implemented start and end dates
- changed shoutouts to times clicked
- updated testing and documentation to reflect changes

## Changes Made

- added endDate, startDate to Flyer model, removed date field
- updated shoutouts field in Flyer model to times clicked
- made optional fields in Organization model nullable
- created queries getFlyersAfterDate() and getFlyersBeforeDate()
- added unit test cases
- updated documentation

## Test Coverage

- jest testing runs to completion
- tested with ios frontend

### DIFF
--- a/src/entities/Flyer.ts
+++ b/src/entities/Flyer.ts
@@ -9,7 +9,7 @@ export class Flyer {
 
   @Field()
   @Property()
-  date: Date;
+  endDate: Date;
 
   @Field()
   @Property({ nullable: true })
@@ -40,8 +40,12 @@ export class Flyer {
   organizationSlugs: [string];
 
   @Field()
+  @Property()
+  startDate: Date;
+
+  @Field()
   @Property({ default: 0 })
-  shoutouts?: number;
+  timesClicked?: number;
 
   @Field()
   @Property()
--- a/src/repos/FlyerRepo.ts
+++ b/src/repos/FlyerRepo.ts
@@ -31,7 +31,7 @@ const getAllFlyers = async (offset = DEFAULT_OFFSET, limit = DEFAULT_LIMIT): Pro
     });
 };
 
-const getFlyersAfterDate = async (since: string, limit = DEFAULT_LIMIT): Promise<Flyer[]> => {
+const getFlyersAfterDate = (since: string, limit = DEFAULT_LIMIT): Promise<Flyer[]> => {
   return (
     FlyerModel.find({
       // Get all Flyers after or on the desired date
@@ -46,7 +46,7 @@ const getFlyersAfterDate = async (since: string, limit = DEFAULT_LIMIT): Promise
   );
 };
 
-const getFlyersBeforeDate = async (before: string, limit = DEFAULT_LIMIT): Promise<Flyer[]> => {
+const getFlyersBeforeDate = (before: string, limit = DEFAULT_LIMIT): Promise<Flyer[]> => {
   return (
     FlyerModel.find({
       // Get all Flyers before the desired date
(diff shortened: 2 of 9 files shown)


src/entities/Flyer.ts
```diff
@@ -39,9 +39,13 @@ export class Flyer {
...
   organizationSlugs: [string];
 
+  @Field()
+  @Property()
+  startDate: Date;
+
   @Field()
   @Property({ default: 0 })
-  shoutouts?: number;
+  timesClicked?: number;
```
Archit404Error, line 48: Are we getting rid of shoutouts and just keeping track of the times clicked instead? In the future document something like this in the PR description -- i.e., "Removed shoutouts in lieu of a timesClicked field because we've chosen not to implement shoutouts for flyers."

src/repos/FlyerRepo.ts
```diff
@@ -23,14 +23,44 @@ function isFlyerFiltered(flyer: Flyer) {
...
+      endDate: { $gte: new Date(since) },
+    })
+      // Sort dates in order of most recent to least
+      .sort({ endDate: 'desc' })
+      .limit(limit)
+      .then((flyers) => {
+        return flyers.filter((flyer) => flyer !== null && !isFlyerFiltered(flyer));
+      })
+  );
+};
```
Archit404Error, line 47: It's generally a bad idea to mix `async` and `.then` syntax in JS/TS. Either change this to use an `await` or remove the `async` at the start of the function definition.

src/repos/FlyerRepo.ts
```diff
@@ -23,14 +23,44 @@ function isFlyerFiltered(flyer: Flyer) {
...
+
```
Archit404Error, line 63: Same `await` and `.then` issue as above

src/tests/data/FactoryUtils.ts
```diff
@@ -25,5 +25,27 @@ class FactoryUtils {
   public static compareByDate(a, b) {
     return new Date(b.date).getTime() - new Date(a.date).getTime();
   }
+
+  /**
+   * Compares the time between two end dates by most recent first
+   *
+   * @param a The first date to be compared
+   * @param b The second date to be compared
+   * @returns 1 if b is more recent than a, -1 if a is more recent than b, 0 if a and b are at the same time
```
Archit404Error, line 34: The `@returns` description here should either state that you return a positive/negative number based on the recentness of the given dates or you should change your function to return +/- 1 (currently you're just returning the difference between the dates in ms).

src/tests/data/FactoryUtils.ts
```diff
@@ -25,5 +25,27 @@ class FactoryUtils {
...
+  public static compareByEndDate(a, b) {
+    return new Date(b.endDate).getTime() - new Date(a.endDate).getTime();
+  }
+
+  /**
+   * Compares the time between two start dates by most recent first
+   *
+   * @param a The first date to be compared
+   * @param b The second date to be compared
+   * @returns 1 if b is more recent than a, -1 if a is more recent than b, 0 if a and b are at the same time
```
Archit404Error, line 45: Same as above

Archit404Error (dismissed):
Good to go. A few small stylistic changes but otherwise good work!

Archit404Error (approved):
Good to go. In the future be sure to resolve all the comments I left by clicking the "resolve conversation" button (I've done it for you in this PR) once you've pushed changes that address those comments so that way re-reviewing is easier.

Archit404Error (approved):
Good to go!

PR #95, opened 2023-06-11
Implement trending flyers logic

isaachan100:
## Overview

- implement trending flyers logic (previously was random sample of flyers)

## Changes Made

- updated trending flyers logic to reflect the number of clicks a flyer has received weighted by how close the event is to occuring
- updated FlyerRepo to reflect changes in logic
- updated documentation in FlyerResolver
- updated constants to include the maximum number of trending flyers

## Test Coverage

- runs locally
- jest testing runs to completion

### DIFF
--- a/src/repos/FlyerRepo.ts
+++ b/src/repos/FlyerRepo.ts
@@ -8,6 +8,7 @@ import {
   DEFAULT_OFFSET,
   FILTERED_WORDS,
   MAX_NUM_DAYS_OF_TRENDING_ARTICLES,
+  MAX_NUM_OF_TRENDING_FLYERS,
 } from '../common/constants';
 import { OrganizationModel } from '../entities/Organization';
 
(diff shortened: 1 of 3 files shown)


src/repos/FlyerRepo.ts
```diff
@@ -178,17 +179,19 @@ const refreshTrendingFlyers = async (): Promise<Flyer[]> => {
...
+      .sort({ trendiness: 'desc' })
+      // Only get flyers for events that start in the next few days
+      .match({
+        startDate: {
+          $lte: new Date(
+            new Date().setDate(new Date().getDate() + MAX_NUM_DAYS_OF_TRENDING_ARTICLES),
+          ),
+        },
+      })
+  ).slice(0, MAX_NUM_OF_TRENDING_FLYERS);
```
Archit404Error, line 194: I'm not sure exactly how this resolves but I don't think you need the double `await` here, you should only need the one outside the parenthesis in order to do what you want

Archit404Error (approved):
Good to go, just double check your double await below

PR #99, opened 2023-08-31
Flyer Notifications + Org/Flyer Model Changes

vinnie4k:
## Overview
This PR implements flyers notification functionality as well as new changes to the Flyer and Organization models.

### Flyer Model

- Added `categorySlug` field
- Made changes to how optional fields are defined to follow the same format as other models such as `Publication`
- Replaced `organizations` and `organizationSlugs` field with `organization` and `organizationSlug`
- Removed `mostRecentFlyer` and `numFlyers` FieldResolvers as requested by Jennifer
- Removed `nsfw` field in the Flyer model

### Organization Model

- Made changes to how optional fields are defined to follow the same format as other models such as `Publication`
- Made `websiteURL` required as requested by Jennifer

### Flyer Queries

- Rewrote `getFlyersByOrganizationSlug` and `getFlyersByOrganizationSlugs` queries in response to changing the Flyer model to only store a single organization

### Flyer Notifications

- Created a POST endpoint for sending a push notification for flyers
- Created `notifyNewFlyers` and `sendNewFlyerNotification` that sends a push notification to followers via FCM when the organization uploads a new flyer

### Other Changes

- Removed `pull_request` from the prod YAML file causing a new Docker image to be built upon creation of a pull request to `release`
- Rewrote test cases in response to new changes in the schema
- Removed `sendNewFlyerNotification`, `sendNewArticleNotification`, `sendNewMagazineNotification`, and `sendWeeklyDebriefNotification`. Moved these calls to be inline with their corresponding notification function.

## Test Coverage
First, I tested to see if the `followOrganization` mutation worked properly. I have a local MongoDB that contains only one user which is my personal iOS device. I went to the GraphQL playground on my local server and tested the mutation with that single user ID. The database seems to properly update so I tested the POST request and received the push notification. See screenshot below. Note that because there is a new schema, I had to rewrite the iOS code to accommodate these changes.

I then had to rewrite a couple of queries and testing factory in response to the new schema. After failing a couple of times, all test cases finally pass .See screenshot below.

One thing to note that it’s very convenient for me to test the new backend changes since I can directly test it with the frontend.

Update: Making the notification calls inline has been tested and everything works properly.

## Next Steps
- Work on implementing the new mutations for creating, editing, and deleting flyers.
- I am also wrapping up on rewriting the environment variables and creating a Volume backend onboarding document for future Volume backend newbies to reference.

## Screenshots
Screenshot 2023-08-30 at 11 24 46 PM
Screenshot 2023-08-30 at 11 01 35 PM

### DIFF
--- a/src/entities/Flyer.ts
+++ b/src/entities/Flyer.ts
@@ -9,11 +9,15 @@ export class Flyer {
 
   @Field()
   @Property()
-  endDate: Date;
+  categorySlug: string;
 
   @Field()
-  @Property({ nullable: true })
-  flyerURL: string;
+  @Property()
+  endDate: Date;
+
+  @Field({ nullable: true })
+  @Property()
+  flyerURL?: string;
 
   @Field()
   @Property()
@@ -27,17 +31,13 @@ export class Flyer {
   @Property()
   location: string;
 
+  @Field((type) => Organization)
+  @Property({ type: () => Organization })
+  organization: Organization;
+
   @Field()
-  @Property({ default: false })
-  nsfw: boolean;
-
-  @Field((type) => [Organization])
-  @Property({ type: () => [Organization] })
-  organizations: [Organization];
-
-  @Field((type) => [String])
-  @Property({ type: () => [String] })
-  organizationSlugs: [string];
+  @Property()
+  organizationSlug: string;
 
   @Field()
   @Property()
--- a/src/repos/NotificationRepo.ts
+++ b/src/repos/NotificationRepo.ts
@@ -1,9 +1,4 @@
 import * as admin from 'firebase-admin';
-import { Article } from '../entities/Article';
-import { Flyer } from '../entities/Flyer';
-import { Magazine } from '../entities/Magazine';
-import { Organization } from '../entities/Organization';
-import { Publication } from '../entities/Publication';
 import { User } from '../entities/User';
 import ArticleRepo from './ArticleRepo';
 import FlyerRepo from './FlyerRepo';
@@ -76,22 +71,6 @@ const sendNotif = async (
   }
 };
 
-const sendNewArticleNotification = async (
-  user: User,
-  article: Article,
-  publication: Publication,
-): Promise<void> => {
-  const notifTitle = publication.name;
-  const notifBody = article.title;
-
-  const uniqueData = {
-    articleID: article.id,
-    articleURL: article.articleURL,
-  };
-
-  sendNotif(user, notifTitle, notifBody, uniqueData, 'new_article');
-};
-
 /**
  * Send notifications for new articles have been posted by publications.
  */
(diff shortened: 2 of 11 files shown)


src/entities/Flyer.ts
```diff
@@ -27,17 +31,13 @@ export class Flyer {
   @Property()
   location: string;
 
-  @Field()
-  @Property({ default: false })
-  nsfw: boolean;
```
Archit404Error, line 32: Seems like you removed nsfw field but didn't document this in the PR description, in the future remember to document all changes

src/repos/NotificationRepo.ts
```diff
@@ -132,6 +136,36 @@ const notifyNewMagazines = async (magazineIDs: string[]): Promise<void> => {
...
+/**
+ * Send notifications for new flyers posted by organizations.
+ */
+const notifyNewFlyers = async (flyerIDs: string[]): Promise<void> => {
+  flyerIDs.forEach(async (f) => {
+    const flyer = await FlyerRepo.getFlyerByID(f); // eslint-disable-line
+    const organization = await OrganizationRepo.getOrganizationBySlug(flyer.organizationSlug);
+    const followers = await UserRepo.getUsersFollowingOrganization(flyer.organizationSlug);
+    followers.forEach(async (follower) => {
+      sendNewFlyerNotification(follower, flyer, organization);
```
Archit404Error, line 164: I don't see a good reason to factor out this logic into a function unless you plan to expose `sendNewFlyerNotification` like the other functions `sendNewMagazineNotification`, etc. are. I know that we also don't use those anywhere else in the codebase, but it seems like we might have been planning to at some point or something.

For now I'd say you can just put the logic of `sendNewFlyerNotification` inside the ForEach and delete the additional function, since it makes the logic harder to follow

Archit404Error (commented):
Essentially good to go, just take a look at comments below. Also, as advice for the future I'd suggest making the PR as modular as possible -- I know this is a relatively small PR but it would've been even easier to review if you had split it into 2 PRs, one for adding notifs for flyers and the other for revamping the models you edited.

Once you address my comment about the function that can be inlined I'll approve.

Nice work on the PR description though, I like the detail! You should also rename the PR title to something more descriptive though.

src/entities/Flyer.ts
vinnie4k: Ah thank you. I knew I was forgetting something.

src/repos/NotificationRepo.ts
vinnie4k: Looking at the code, it seems like `sendNewMagazineNotification` and `sendNewArticleNotification` are exported but not actually used anywhere. Should we put this logic in their respective `forEach` or do we just keep it as is? Whatever approach we use is fine as long as we keep it consistent.

Archit404Error: Yeah let's just inline all of those functions then

Archit404Error (approved):
Good to go once Joyce approves

joycrwu (approved):
lgtm! thanks archit for the pr specific comments

PR #100, opened 2023-09-01
getFlyersByCategorySlug Query

vinnie4k:
## Overview
This PR implements a query that returns flyers given a category slug.

### getFlyersByCategorySlug
- Created a query that returns a list of flyers of size `limit` (default 25) given a `categorySlug`, offset by `offset` (default 0). These flyers are sorted by start date descending.
- Wrote various test cases.

### Other Changes
- Removed `checkProfanity` field resolver for Flyers

## Test Coverage
- Wrote test cases that cover no flyers, one flyer returned with multiple flyers stored, and flyers of size `limit`. See screenshot below for passed test cases.
- I also tested with my other local database through the GraphQL playground to see if the flyers were sorted properly as well as testing `limit` and `offset` parameters. See screenshot below.

## Next Steps
- Work on implementing the new mutations for creating, editing, and deleting flyers.

## Screenshots
PNG image
getFlyersByCategoryPlayground

### DIFF
--- a/src/repos/FlyerRepo.ts
+++ b/src/repos/FlyerRepo.ts
@@ -84,7 +84,7 @@ const getFlyersByOrganizationSlug = async (
   limit: number = DEFAULT_LIMIT,
   offset: number = DEFAULT_OFFSET,
 ): Promise<Flyer[]> => {
-  return FlyerModel.find({ organizationSlugs: slug })
+  return FlyerModel.find({ 'organization.slug': slug })
     .sort({ startDate: 'desc' })
     .skip(offset)
     .limit(limit)
@@ -99,7 +99,7 @@ const getFlyersByOrganizationSlugs = async (
   offset: number = DEFAULT_OFFSET,
 ): Promise<Flyer[]> => {
   const uniqueSlugs = [...new Set(slugs)];
-  return FlyerModel.find({ organizationSlugs: { $in: uniqueSlugs } })
+  return FlyerModel.find({ 'organization.slug': { $in: uniqueSlugs } })
     .sort({ startDate: 'desc' })
     .skip(offset)
     .limit(limit)
--- a/src/repos/FlyerRepo.ts
+++ b/src/repos/FlyerRepo.ts
@@ -8,6 +8,7 @@ import {
   DEFAULT_OFFSET,
   FILTERED_WORDS,
   MAX_NUM_DAYS_OF_TRENDING_ARTICLES,
+  MAX_NUM_OF_TRENDING_FLYERS,
 } from '../common/constants';
 import { OrganizationModel } from '../entities/Organization';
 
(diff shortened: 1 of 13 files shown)


src/repos/FlyerRepo.ts
```diff
@@ -62,6 +62,20 @@ const getFlyersBeforeDate = (before: string, limit = DEFAULT_LIMIT): Promise<Fly
   );
 };
 
+const getFlyersByCategorySlug = async (
+  categorySlug: string,
+  limit = DEFAULT_LIMIT,
+  offset = DEFAULT_OFFSET,
+): Promise<Flyer[]> => {
+  return FlyerModel.find({ categorySlug })
```
Archit404Error, line 70: I know this is syntactic sugar but let's make it more explicit, so that we'll be able to update the query appropriately if we ever change the field name to something else (I'm not sure whether the current code will throw a compile time error if the fields are updated): 

```suggestion
  return FlyerModel.find({ categorySlug: categorySlug })
```

src/tests/flyer.test.ts
```diff
@@ -202,4 +202,43 @@ describe('getTrending tests', () => {
     const getFlyersResponse = await FlyerRepo.getTrendingFlyers();
     expect(getFlyersResponse).toHaveLength(5);
   });
+
+  describe('getFlyersByCategorySlug tests', () => {
+    test('getFlyersByCategorySlug - no flyers', async () => {
```
Archit404Error, line 207: Let's update this test name to be more descriptive -- something like "invalid slug" should be contained in the name since we're testing how the function deals with invalid slugs

src/tests/flyer.test.ts
```diff
@@ -202,4 +202,43 @@ describe('getTrending tests', () => {
...
+      const flyers = await FlyerFactory.create(2);
+      await FlyerModel.insertMany(flyers);
+
+      const getFlyersResponse = await FlyerRepo.getFlyersByCategorySlug(Math.random().toString());
+      expect(getFlyersResponse).toHaveLength(0);
+    });
+
+    test('getFlyersByCategorySlug - 1 flyer returned, multiple flyers stored', async () => {
```
Archit404Error, line 215: Once again, let's make this more descriptive -- here we're trying to find a flyer by slug after inserting a bunch of flyers with only one of them having the expected slug, so maybe something like "getFlyersByCategorySlug - query flyer with existing slug"

src/tests/flyer.test.ts
```diff
@@ -202,4 +202,43 @@ describe('getTrending tests', () => {
...
+      const randomSlug = Math.random().toString();
+      const specificFlyer = await FlyerFactory.createSpecific(1, { categorySlug: randomSlug });
+      await FlyerModel.insertMany(specificFlyer);
+
+      const getFlyersResponse = await FlyerRepo.getFlyersByCategorySlug(randomSlug);
+      expect(getFlyersResponse[0].categorySlug).toEqual(specificFlyer[0].categorySlug);
+      expect(getFlyersResponse).toHaveLength(1);
+    });
+
+    test('getFlyersByCategorySlug - 2 flyers returned, restrict to limit', async () => {
```
Archit404Error, line 228: Same comment here about making this more descriptive

Archit404Error (approved):
Good to go with a few small comments:
- I'd suggest forgoing the sugar and being more explicit with the find in case the field name every changes -- I don't believe a compile time error would be thrown if we used the sugar (and even if it was it'd be harder to figure out what's wrong)
- Some comments about the test suite -- I know you made the tests in accordance with how the other flyer tests were written but I'm not a fan of the way the tests have been named in this codebase. The `describe()` already separates our tests out so we shouldn't need to prefix every test with the section/function name, so each individual test case should just be describing what the test is doing, not the return type or number of flyers returned
- At some point we should refactor the tests in this codebase

src/repos/FlyerRepo.ts
vinnie4k: ESLint wants the shorthand notation...

Archit404Error (approved):
Good to go

PR #102, opened 2023-09-06
`createFlyer` mutation

vinnie4k:
## Overview

This PR implements a mutation to create a flyer.

### createFlyer Mutation

- Created a mutation that returns a newly created Flyer if successful given the following arguments:
    - {string} `categorySlug` the slug for this flyer's category
    - {string} `endDate` the end date for this flyer's event in UTC ISO8601 format
    - {string} `flyerURL` the URL for this flyer when tapped
    - {string} `imageB64` the base64 string representing the flyer's image
    - {string} `location` the location for this flyer's event
    - {string} `organizationID` the ID of the organization creating this flyer
    - {string} `startDate` the start date for this flyer's event in UTC ISO8601 format
    - {string} `title` the title for this flyer
- The `imageB64` is uploaded to our Digital Ocean via a POST request.

### FlyerErrorInterceptor

- Wrote a middleware that attempts to upload an image to our servers. It will return via context the image URL if successful; otherwise, an error is thrown.
- The purpose of this middleware is to prevent a new Flyer from being created if the upload request fails.

### uploadImage

- Wrote a function that uploads an image given a base64 string and returns the generated image URL.

## Test Coverage
- I started by testing this mutation through the GraphQL playground with my local MongoDB:

```
mutation CreateFlyer($title: String!, $startDate: String!, $organizationID: String!, $location: String!, $imageB64: String!, $flyerURL: String, $endDate: String!, $categorySlug: String!) {
  createFlyer(title: $title, startDate: $startDate, organizationID: $organizationID, location: $location, imageB64: $imageB64, flyerURL: $flyerURL, endDate: $endDate, categorySlug: $categorySlug) {
    ...flyerFields
  }
}
```

- To ensure that this mutation works as expected, I provided the following invalid arguments to see if the request fails:
    - Invalid `organizationID`
    - Invalid `startDate` and `endDate` format (note that the frontend must provide this in the correct format and timezone; the backend will not do any timezone conversion)
        - Example valid format: `YYYY-mm-ddTHH:MM:ssZ`
        - This is required in order to store the dates as Date objects in MongoDB.
    - Empty `flyerURL` since this is optional
    - Invalid `imageB64`
        - An error is thrown by GraphQL if the upload request fails. See second screenshot below.
- See the first screenshot below for a successful request.

## Next Steps
- Work on implementing the editing and deleting flyers mutations.
- I believe that all functions should have proper documentation and the test suite description should also be refactored. Perhaps this could be something to do in the future.

## Screenshots
[...]

PR #103, opened 2023-09-06
`deleteFlyer` mutation

vinnie4k:
## Overview

This PR implements a mutation to delete a flyer.

### deleteFlyer Mutation

- Created a mutation that deletes a Flyer if successful given the following arguments:
    - {string} `id` the ID of the Flyer to delete
- This mutation returns the deleted flyer. If the flyer cannot be found, the mutation will throw an error.
- **Note that if the POST request to remove the flyer image from our servers fails, the Flyer will still be deleted in the MongoDB.**

### removeImage

- Created a function in `utils.ts` to remove an image from our servers, given the bucket name and image URL
- Created an interface `UploadResponse` with the properties `data: String` and `success: boolean` to store the parsed JSON response.

## Test Coverage
- I started by testing this mutation through the GraphQL playground with my local MongoDB:

```
mutation DeleteFlyer($id: String!) {
  deleteFlyer(id: $id) {
    ...flyerFields
  }
}
```

- To ensure that this mutation works as expected, I provided the following invalid arguments to see if the request fails:
    - Invalid `id`. See the first screenshot.
- The second screenshot shows a successful delete mutation. My local MongoDB is updated correctly.
- I then wrote test cases to test this mutation. See screenshot below.

## Next Steps
- Work on implementing editing flyer mutation.
- I believe that all functions should have proper documentation and the test suite description should also be refactored. Perhaps this could be something to do in the future.

## Screenshots
delete_flyer_1
delete_flyer_2
delete_flyer_3
delete_flyer_4

### DIFF
--- a/src/repos/FlyerRepo.ts
+++ b/src/repos/FlyerRepo.ts
@@ -238,7 +238,6 @@ const incrementTimesClicked = async (id: string): Promise<Flyer> => {
 /**
  * Create a new flyer.
  *
- * @function
  * @param {string} categorySlug the slug for this flyer's category
  * @param {string} endDate the end date for this flyer's event in UTC ISO8601 format
  * @param {string} flyerURL the URL for this flyer when tapped
@@ -281,7 +280,6 @@ const createFlyer = async (
 /**
  * Delete a flyer
  *
- * @function
  * @param id the flyer ID to remove
  */
 const deleteFlyer = async (id: string): Promise<Flyer> => {
--- a/src/repos/FlyerRepo.ts
+++ b/src/repos/FlyerRepo.ts
@@ -11,6 +11,7 @@ import {
   MAX_NUM_OF_TRENDING_FLYERS,
 } from '../common/constants';
 import { OrganizationModel } from '../entities/Organization';
+import utils from '../utils';
 
 const { IS_FILTER_ACTIVE } = process.env;
 
@@ -218,6 +219,7 @@ const refreshTrendingFlyers = async (): Promise<Flyer[]> => {
 
 /**
  * Increments number of times clicked on a flyer by one.
+ *
  * @function
  * @param {string} id - string representing the unique Object Id of a flyer.
  */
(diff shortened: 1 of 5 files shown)


src/common/constants.ts
```diff
@@ -20,4 +20,4 @@ export const MAX_NUM_DAYS_OF_TRENDING_ARTICLES = 30;
 export const MAX_NUM_DAYS_OF_FEATURED_MAGAZINES = 30;
 export const MAX_NUM_OF_TRENDING_FLYERS = 10;
 export const UPLOAD_BUCKET = 'volume';
-export const UPLOAD_SERVICE_URL = 'https://upload.cornellappdev.com/upload/';
+export const UPLOAD_SERVICE_URL = 'https://upload.cornellappdev.com';
```
Archit404Error, line 23: This needs to be removed form the git history as well in accordance with #102, you should try to git reset your commits here and force a rebase onto your other branch once you've removed the sensitive information there. Will come back to this and do a full review once that's done.

Archit404Error (commented):
Just suggesting this one change for now, will re-review once this is resolved

src/repos/FlyerRepo.ts
```diff
@@ -275,8 +278,26 @@ const createFlyer = async (
   return FlyerModel.create(newFlyer);
 };
 
+/**
+ * Delete a flyer
+ *
+ * @function
```
Archit404Error, line 284: Generally the `@function` is used in JSDoc when we want to specify an object is a function if it is unclear (e.g., if we did `let x = someFunc()` and someFunc() returns a function, we would mark x with @function. I don't think we need this here

src/utils.ts
```diff
@@ -22,4 +22,28 @@ const uploadImage = async (imageB64: string): Promise<string> => {
...
+  const payload = {
+    bucket: UPLOAD_BUCKET,
+    image_url: imageURL,
+  };
+  const response = await fetch(`${process.env.UPLOAD_SERVICE_URL}/remove/`, {
+    method: 'POST',
+    body: JSON.stringify(payload),
+  });
+
+  const responseData = JSON.parse(await response.text());
```
Archit404Error, line 42:
```suggestion
  const responseData = await response.json();
```

src/utils.ts
```diff
@@ -22,4 +22,28 @@ const uploadImage = async (imageB64: string): Promise<string> => {
...
+/**
+ * Remove an image from AppDev servers
+ *
+ * @param imageURL the image URL to remove
+ * @returns true if the deletion was successful; otherwise false
+ */
+const removeImage = async (imageURL: string): Promise<boolean> => {
+  // Delete image via a POST request
+  const payload = {
+    bucket: UPLOAD_BUCKET,
```
Archit404Error, line 34: Once again would recommend moving this upload bucket variable into the env file as well and purging from git history

Archit404Error (changes requested):
Essentially good to go, I want you to move the upload bucket into the env file in #102 and then rebase this branch onto that one, purging the upload bucket from the constants file as well since it's good to keep the buckets secret.

Once that's done I'll approve

src/utils.ts
vinnie4k: I tried doing this before but the issue is that I want to access the `success` propery of the JSON. Using `response.json()` will give `responseData` an unknown type so Typescript will say "Property `success` does not exist on type `unknown`".

Archit404Error: Yes what you should do then is define a type that specifies what the response JSON will look like and specify that that is the type of the JSON response data. You can checkout an example of this in one of my repos. I define the type interface [here](https://github.com/Archit404Error/SemanticClassSearch/blob/main/mobile/index.d.ts) and make the request [here](https://github.com/Archit404Error/SemanticClassSearch/blob/8d1db8c34490bfbe428080ff4a43802640e54d5c/mobile/app/(tabs)/search/index.tsx#L9-L14)

vinnie4k: I created an interface at the top of `utils.ts` called `UploadResponse` and wrote the following:
```
const responseData = (await response.json()) as UploadResponse;
if (!responseData.success) {
  console.log(`Removing an image from our servers failed with image URL: ${imageURL}`);
}
return responseData.success;
```
This works as intended, but do you want to have a `.d.ts` file instead?

Archit404Error: The interface should be fine for now, we can always pull this out into a .d.ts file later

Archit404Error: You might also be able to just do `const responseData : UploadResponse = await response.json()`. Once you commit this I'll approve

Archit404Error (commented):
Will approve once you add in the interface, try doing `const resJson : Interface = await response.json()` and if that doesn't work you can use the existing code you have

src/utils.ts
vinnie4k: `const responseData : UploadResponse = await response.json()` doesn't work. I believe it's because the response has an `undefined` type. I kept what I wrote earlier and everything works perfectly for adding and deleting flyers.

src/repos/FlyerRepo.ts
```diff
@@ -218,6 +219,7 @@ const refreshTrendingFlyers = async (): Promise<Flyer[]> => {
 
 /**
  * Increments number of times clicked on a flyer by one.
+ *
  * @function
```
Archit404Error, line 223: Remove this

Archit404Error (approved):
Good to go, just remove the extraneous `@function` in your JSDoc

PR #106, opened 2023-09-14
Trendiness Update and Migration Script for New Flyers Model

zachseidner1:
## Overview

This PR updates the `getTrendingFlyers` endpoint to use a simpler query that is simply based on the trendiness of the Flyer. The `isTrending` parameter was removed in favor of sorting the Flyers by `trendiness` and returning the top trending Flyers that are not out of date. This PR also includes a migration script to port the data from the old flyers model to the new flyers model. Specifically, going from the array of organizations to a single organization object.

## Changes Made

- Changed `getTrendingFlyers` endpoint to sort by non-outdated Flyers with the highest `trendiness`.
- Rewrote unit tests for `getTrendingFlyers` to test new route.
- Create migration script to migrate to the new Flyers model
- Move previous migration script to `past-migrations` folder.
- Remove `isTrending` parameter from Flyers model.

## Test Coverage

Rewrote unit tests to test new `getTrendingFlyers` route. One unit test ensures that there are no outdated flyers regardless of trendiness. The other ensures that the route gets the flyers with highest `trendiness` first.

## Related PRs 

Addresses the flyer model changes from #99 

## Screenshots 

image


src/repos/FlyerRepo.ts
```diff
@@ -166,45 +160,18 @@ const searchFlyers = async (query: string, limit = DEFAULT_LIMIT) => {
...
-  ).slice(0, MAX_NUM_OF_TRENDING_FLYERS);
-
-  flyers.forEach(async (a) => {
-    const flyer = await FlyerModel.findById(new ObjectId(a._id)); // eslint-disable-line
-    flyer.isTrending = true;
-    await flyer.save();
-  });
-
-  return flyers;
+      .then((flyers) => flyers.slice(0, limit))
```
Archit404Error, line 173: Generally not a good idea to use `.then` inside an `async` function, you should always pair async functions with await syntax. Also, in this case you can just use the built in `limit()` method mongoose provides (see change suggestion below)

```suggestion
      .limit(limit)
```

I looked through the codebase and saw that this anti pattern of calling .then in async repo functions is common -- eventually this should be fixed

src/tests/flyer.test.ts
```diff
@@ -191,16 +191,57 @@ describe('incrementShoutouts tests', () => {
...
+    // Create flyers with random trendiness
+    const randomFlyers = await FlyerFactory.createSpecificsByIndex(
+      randomTrendiness.length,
+      (index) => {
+        return {
+          trendiness: randomTrendiness[index],
+          title: `title${randomTrendiness[index]}`,
+        };
+      },
+    );
```
Archit404Error, line 205: This approach is interesting, but I don't see any major advantage to this call over
```js
const randomFlyers = await FlyerFactory.create(n);
for (let i = 0; i < randomFlyers.length; i++) {
  randomFlyers[i].trendiness = randomTrendiness[index]
  randomFlyers[i].title = `title${randomTrendiness[index]}`;
  // insert any other attributes you want to modify here
}
```

The purpose of writing static methods like this is to re-use code and reduce complexity but in this case the for loop approach takes essentially the same amount of lines (less when you count lines saved by not writing the function).

Even if we write a for loop every time we want to modify the attributes of the random flyers in a test case, it'll take essentially the same amount of work as calling this function. We could technically pull out your anonymous function and try to re-use that but I think this just convolutes the code too much.

I tend to really prefer simplicity of code above all else, so I think we should just go with the straight for loop approach here

src/tests/data/FlyerFactory.ts
```diff
@@ -38,6 +38,31 @@ class FlyerFactory {
...
+     */
+    const arr = await FlyerFactory.create(n);
+    return arr.map((x, i) => {
+      const newDoc = x;
+      Object.entries(newMappings(i)).forEach(([k, v]) => {
+        newDoc[k] = v;
+      });
+      return newDoc;
+    });
+  }
```
Archit404Error, line 64: I currently don't see the value of this function over a for loop that manually modifies attributes, see my comment below

src/tests/flyer.test.ts
```diff
@@ -191,16 +191,57 @@ describe('incrementShoutouts tests', () => {
...
```
Archit404Error, line 228: Once again I'd just replace this with a for loop, see above

Archit404Error (changes requested):
This looks solid so far, take a look at my comments below

src/tests/flyer.test.ts
zachseidner1: I agree - honestly I didn't know that you could just write a for loop, and I was inspired by Kotlin's `forEachIndexed` style of functional programming. But I see how that adds complexity for almost no gain in efficiency, so a for loop works better in this case. Thanks for the tip and I removed the function from the code.

zachseidner1:
@Archit404Error Just addressed your PR comments. Really appreciate the detail and effort you put in, I can tell you're really reading and analyzing my code to help make sure the code base is high quality. Hope now after these changes my PR is good to go.

Archit404Error (approved):
Good to go, great job on the PR and on implementing the fixes!

PR #107, opened 2023-09-14
Zach/trendiness and migration hot fix

zachseidner1:
## Overview
Quick hot fix for a syntax error I left out from #106 . Backend now runs properly. 

image
image


Archit404Error (approved):
Approved - in the future be sure to actually run the backend before making a PR! Your Jest tests will miss scheduled/async code, incorrect function calls, etc., I had assumed you'd already done so before making the previous PR

PR #109, opened 2023-09-26
`checkAccessCode` query for Organization authentication

vinnie4k:
## Overview
Created queries and updated the Organization model to allow for authentication of an organization.

### Organization Model
- Added `accessCode` field (a string) to the `Organization` model.

### Access Code Verification
- Added `bcrypt` package for hashing.
- Created `checkAccessCode` query to validate an organization given a slug and their non-hashed access code.
    - We will be using bcrypt to compare the access code with the hashed password. We are using `.then` syntax (recommended by bcrypt) since asynchronisation splits the crypto operation into small chunks.
    - If validated, this function will return the organization.

## Testing
- Made `accessCode` nullable temporarily in order to test locally without having to do a migration.
- First we tested to see if the `checkAccessCode` function properly throws an error so that it can be handled on the frontend side.
    - If the access code is invalid with the given slug, then an error is throw to be handled on the frontend side.
- Additionally, if the access code does not match, then the value of `result` after the promise is resolved will be `false`.
    - This will throw an `UnauthorizedError` that the frontend can then handle. We are using `UnauthorizedError` provided by TypeGraphQL because the message header makes a lot more sense.
- We are not writing Jest test cases because that will require `organizations.json` to have an `accessCode` for all organizations. Testing with the GraphQL playground is more than enough.

## Next Steps
- Once migration has been done (Zach and I will manually create the access codes), we will need to set the `accessCode` field to NOT be nullable.

### Invalid Slug
invalid_slug

### Invalid Access Code
invalid

### Valid Access Code  + Slug
valid


src/entities/Organization.ts
```diff
@@ -6,6 +6,10 @@ export class Organization {
   @Field(() => ID)
   id: string;
 
+  @Field({ nullable: true })
+  @Property()
+  accessCode: string;
+
```
Archit404Error, line 12: As discussed, remove field decorator here

src/resolvers/OrganizationResolver.ts
```diff
@@ -43,6 +43,14 @@ class OrganizationResolver {
     return OrganizationRepo.getOrganizationBySlug(slug);
   }
 
+  @Query((_return) => Organization, {
+    nullable: true,
+    description: `Returns the <Organization> given a <slug> and <accessCode>.`,
+  })
+  async checkAccessCode(@Arg('accessCode') accessCode: string, @Arg('slug') slug: string) {
+    return OrganizationRepo.checkAccessCode(accessCode, slug);
+  }
```
Archit404Error, line 52: Maybe we can set up some sort of session token auth here as well -- we don't necessarily need to use the graphQL directly to do auth, we can also just create an express endpoint for this that sends over a session token and then create a middleware for the protected queries and mutations

zachseidner1: We could try something like this, but our concern with using session tokens was if someone logs in and then 1 month later their session expires and they forget their password, we want to avoid a situation when an org forgets their password since we don't have any email recovery system set up.

Archit404Error: If that's your goal you can just ensure the decoded token is valid and then if the date has expired issue a new refreshed token

Archit404Error (approved):
Good to go

zachseidner1 (approved):

PR #110, opened 2023-09-27
Implemented  getAllFlyerCategories and tests

cindy-x-liang:
## Overview

Created getAllFlyerCategories which will return all the different flyer categories of the flyers inside the database.

## Changes Made

Create an async function called getAllFlyerCategories in FlyerRepo.ts. Used the mongoose function distinct() to get unique category slugs from each flyer. Created a query for this function in FlyerResolver.ts, and implemented test cases in flyer test.t

## Test Coverage

Test cases test for instances when all flyers have different categories and when flyers share the same category.

<details>

  <summary>GraphQL Screenshot</summary>

Screenshot 2023-09-27 at 6 00 21 PM

  

</details>


src/repos/FlyerRepo.ts
```diff
@@ -70,6 +70,18 @@ const getFlyersByCategorySlug = async (
...
+/**
+ * Looks for unique category slugs for all the Flyers
+ * @returns a list of Strings with all the category names. 
+ */
+const getAllFlyerCategories = async (
+): Promise<String[]> => {
+  return FlyerModel.collection.distinct("categorySlug")
+    .then((categories) => {
+      return categories;
+    });
```
zachseidner1, line 82: The `.then` method is not needed here. When the `FlyerModel.collection.distinct` call finishes, the `.then` returns a new promise, whose behavior is defined based on the function you put in. But your function just returns the resulting value, so it is redundant. See [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then#return_value) for more info.

zachseidner1 (changes requested):
Nice job! Just see my comment about the `.then` statement.

src/repos/FlyerRepo.ts
vinnie4k, line 82: Do you have Prettier installed? ESLint should also point out formatting errors. Replace this function with:
```
/**
 * Looks for unique category slugs for all the Flyers
 *
 * @returns a list of Strings with all the category names.
 */
const getAllFlyerCategories = async (): Promise<string[]> => {
  return FlyerModel.collection.distinct('categorySlug').then((categories) => {
    return categories;
  });
};
```
Then, address @zachseidner1 's comment on removing `.then`.

src/repos/FlyerRepo.ts
```diff
@@ -269,6 +281,7 @@ export default {
   getFlyersAfterDate,
   getFlyersBeforeDate,
   getFlyersByCategorySlug,
+  getAllFlyerCategories,
```
vinnie4k, line 284: This should be sorted alphabetically.

src/resolvers/FlyerResolver.ts
```diff
@@ -151,6 +151,16 @@ class FlyerResolver {
...
   }
 
+  @Query((_return) => [String], {
+    nullable: false,
+    description: `Returns a list of <Strings> representing all of the categories of the different flyers in this database`,
+  })
+  async getAllFlyerCategories(
+  ) {
+    return FlyerRepo.getAllFlyerCategories();
+  }
```
vinnie4k, line 161: Same as above, replace this line with:
```
@Query((_return) => [String], {
  nullable: false,
  description: `Returns a list of <Strings> representing all of the categories of the different flyers in this database`,
})
async getAllFlyerCategories() {
  return FlyerRepo.getAllFlyerCategories();
}
```

src/resolvers/FlyerResolver.ts
```diff
@@ -151,6 +151,16 @@ class FlyerResolver {
...
+
+
```
vinnie4k, line 163: Remove this line.

src/tests/flyer.test.ts
```diff
@@ -276,6 +276,45 @@ describe('getFlyersByCategorySlug tests', () => {
...
+
+    const randomSlug3 = Math.random().toString();
+    const specificFlyer3 = await FlyerFactory.createSpecific(1, { categorySlug: randomSlug3 });
+    await FlyerModel.insertMany(specificFlyer3);
+
+    const getFlyersResponse = await FlyerRepo.getAllFlyerCategories();
+    expect(getFlyersResponse.sort()).toEqual([randomSlug1, randomSlug2, randomSlug3].sort());
+  });
+
+
```
vinnie4k, line 315: Remove this line.

src/tests/flyer.test.ts
```diff
@@ -276,6 +276,45 @@ describe('getFlyersByCategorySlug tests', () => {
   });
 });
 
+describe('getAllFlyerCategories tests', () => {
```
vinnie4k, line 279: I would put all of this at the end of the file just to keep things in order of implementation but not a big deal.

vinnie4k (changes requested):
Great work! I am very impressed in your ability to see the common patterns in our codebase and being able to learn from them in a short amount of time, especially since you don't have a background in TypeScript.

Please address my comments about formatting and address @zachseidner1 's comments. Then request me again. I would also wait for @Archit404Error the goat to review before merging.

src/repos/FlyerRepo.ts
```diff
@@ -70,6 +70,18 @@ const getFlyersByCategorySlug = async (
     });
 };
 
+/**
+ * Looks for unique category slugs for all the Flyers
+ * @returns a list of Strings with all the category names. 
+ */
+const getAllFlyerCategories = async (
```
Archit404Error, line 77: This function actually doesn't need to be `async`, since all you're doing here is returning a promise, which should actually be a synchronous action. Async/await stuff can be kind of confusing -- as a general rule of thumb, only mark a function `async` if you're using the keyword `await` inside of it!

src/tests/flyer.test.ts
```diff
@@ -276,6 +276,45 @@ describe('getFlyersByCategorySlug tests', () => {
...
+    const randomSlug1 = "Academic";
+    const specificFlyer1 = await FlyerFactory.createSpecific(1, { categorySlug: randomSlug1 });
+    await FlyerModel.insertMany(specificFlyer1);
+
+    const randomSlug2 = "Dance";
+    const specificFlyer2 = await FlyerFactory.createSpecific(1, { categorySlug: randomSlug2 });
+    await FlyerModel.insertMany(specificFlyer2);
+
+    const specificFlyer3 = await FlyerFactory.createSpecific(1, { categorySlug: randomSlug2 });
+    await FlyerModel.insertMany(specificFlyer3);
```
Archit404Error, line 291: A more idiomatic way to write this code would be use an array and leverage `createSpecfic()`'s ability to create more than one of a type of flyer, especially since `insertMany()` allows you to insert multiple documents at once:
```suggestion
    const flyerSlugs = ["Academic", "Dance"];
    let generatedFlyers = [];
    generatedFlyers.concat(await FlyerFactory.createSpecific(1, {categorySlug: flyerSlugs[0]}))
    generatedFlyers.concat(await FlyerFactory.createSpecific(2, {categorySlug: flyerSlugs[1]}))
    
    await FlyerModel.insertMany(generatedFlyers)
```

The code is now a lot shorter/simpler with less variable names to keep track of, while still giving you the same functionality

src/tests/flyer.test.ts
```diff
@@ -276,6 +276,45 @@ describe('getFlyersByCategorySlug tests', () => {
...
+    const specificFlyer1 = await FlyerFactory.createSpecific(1, { categorySlug: randomSlug1 });
+    await FlyerModel.insertMany(specificFlyer1);
+
+    const randomSlug2 = Math.random().toString();
+    const specificFlyer2 = await FlyerFactory.createSpecific(1, { categorySlug: randomSlug2 });
+    await FlyerModel.insertMany(specificFlyer2);
+
+    const randomSlug3 = Math.random().toString();
+    const specificFlyer3 = await FlyerFactory.createSpecific(1, { categorySlug: randomSlug3 });
+    await FlyerModel.insertMany(specificFlyer3);
```
Archit404Error, line 309: I'd suggest rewriting this in the same way I suggested for the test above

src/tests/flyer.test.ts
```diff
@@ -276,6 +276,45 @@ describe('getFlyersByCategorySlug tests', () => {
   });
 });
 
+describe('getAllFlyerCategories tests', () => {
+  test('query 3 different flyers with 2 different categories', async () => {
```
Archit404Error, line 280: Our goal should be to have really descriptive test names -- I know a lot of the other tests in this codebase are structured like the name you've written here, but I'd suggest changing it up to reflect what we're actually testing. Our goal here is to show that our query filters out duplicate slugs, so something like `Handle non-unique CategorySlugs across flyers in collection` or `ensure only unique categorySlugs are returned` would make more sense here

Archit404Error (changes requested):
Great job on your first PR! I think Vin and Zach covered a lot of the formatting details -- once you get your ESLint and auto formatting going you should be able to resolve those quickly.

I wanted to highlight some areas of your code that could be optimized a bit more for readability -- the goal is always to make sure we have idiomatic code that is easy to understand and modify, so you can check those comments out below

Overall, though, great job on this PR!

src/tests/flyer.test.ts
```diff
@@ -294,4 +293,41 @@ describe('deleteFlyer tests', () => {
     const deleteFlyerResponse = await FlyerRepo.deleteFlyer('64811792f910705ca1a981f8');
     expect(deleteFlyerResponse).toBeNull();
   });
-});
\ No newline at end of file
+});
+
+describe('Handle non-unique CategorySlugs across flyers in collection', () => {
```
vinnie4k, line 298: Keep the text inside of `describe` the same as you had before: `describe('getAllFlyerCategories tests', () => {`
What Archit meant was for the `test` part since each test case is testing for something different. At some point, we will need to refactor the entire testing suite and write test cases for resolvers.

vinnie4k (approved):
Fix your test description and you should be good to go on my part. Once @zachseidner1 and @Archit404Error approve, feel free to merge. Also, make sure you mark the comment with "Resolve conversation" so that we know you have resolved it. Great work!

src/tests/flyer.test.ts
```diff
@@ -294,4 +293,41 @@ describe('deleteFlyer tests', () => {
...
+    const generatedFlyers = [];
+    generatedFlyers.push(
+      (await FlyerFactory.createSpecific(1, { categorySlug: flyerSlugs[0] }))[0],
+    );
+    generatedFlyers.push(
+      (await FlyerFactory.createSpecific(1, { categorySlug: flyerSlugs[1] }))[0],
+    );
+    generatedFlyers.push(
+      (await FlyerFactory.createSpecific(1, { categorySlug: flyerSlugs[2] }))[0],
+    );
```
zachseidner1, line 326: Don't repeat yourself! Replace this with a for loop:
```TypeScript
for (const slug of flyerSlugs) {
      generatedFlyers.push((await FlyerFactory.createSpecific(1, { categorySlug: slug }))[0]);
    }
```

src/tests/flyer.test.ts
```diff
@@ -294,4 +293,41 @@ describe('deleteFlyer tests', () => {
...
+describe('getAllFlyerCategories tests', () => {
+  test('Ensure only unique CategorySlugs across flyers in collection are returned', async () => {
+    const flyerSlugs = ['Academic', 'Dance'];
+    const generatedFlyers = [];
+    generatedFlyers.push(
+      (await FlyerFactory.createSpecific(1, { categorySlug: flyerSlugs[0] }))[0],
+    );
+    generatedFlyers.push(
+      (await FlyerFactory.createSpecific(2, { categorySlug: flyerSlugs[1] }))[0],
+    );
```
zachseidner1, line 307: You may want to make this a for loop as well so the tests follow the same style
```TypeScript
for (const slug of flyerSlugs) {
      generatedFlyers.push((await FlyerFactory.createSpecific(1, { categorySlug: slug }))[0]);
    }
```

zachseidner1 (changes requested):
I know my comments tend to be critical, but I do want to emphasize that this is great work for your first PR, especially considering you don't have much experience with TypeScript. Just work a bit on code concision, but other than that awesome work!

src/tests/flyer.test.ts
Archit404Error: I actually wouldn't recommend using a for loop here because the goal is to create 3 flyers, two of which have the same slug. The code above would only test 2 flyers with 2 unique slugs which is semantically what the other test does. You could technically make this a loop by making the categorySlugs array ["Category1", "Category2", "Category2"] but honestly I think the current code is a lot more readable

Archit404Error (dismissed):
Good to go once you resolve all comments above!

src/tests/flyer.test.ts
zachseidner1: Interesting, I would argue that the `["Category1", "Category2", "Category2"]` is more readable, since it clearly outlines which category slugs we are adding to `generatedFlyers` at the top of the function, as opposed to the current code, where one might skip over the `2` in the parameters of the `createSpecific` function (this is what I did while reading it). But we're just nitpicking at this point, it doesn't really matter which approach you use. That's why I said you _may_ want to make this a for loop. But imo it should be up to Cindy at this point.

zachseidner1: Wait just noticed that the current code actually only adds 2 distinct Flyers!
`(await FlyerFactory.createSpecific(2, { categorySlug: flyerSlugs[1] }))[0]`
only adds the Flyer at index 0 of the generated array of 2 Flyers. So she needs to either use the for-loop approach or use `generatedFlyers.concat` and remove the accessor.

Archit404Error: Oh yeah I had assumed this code was using `concat()` like my suggestion above haha.

zachseidner1 (dismissed):

src/tests/flyer.test.ts
```diff
@@ -294,4 +293,32 @@ describe('deleteFlyer tests', () => {
...
-});
\ No newline at end of file
+});
+
+describe('getAllFlyerCategories tests', () => {
+  test('Ensure only unique CategorySlugs across flyers in collection are returned', async () => {
+    const flyerSlugs = ['Academic', 'Dance', 'Dance'];
+    const generatedFlyers: Flyer[] = [];
+    for (const slug of flyerSlugs) {
+      generatedFlyers.push(...(await FlyerFactory.createSpecific(1, { categorySlug: slug })));
```
Archit404Error, line 303: Using concat is cleaner than trying to unpack the array:

```suggestion
      generatedFlyers.concat(await FlyerFactory.createSpecific(1, { categorySlug: slug }));
```

src/tests/flyer.test.ts
```diff
@@ -294,4 +293,32 @@ describe('deleteFlyer tests', () => {
...
+
+    const getFlyersResponse = await FlyerRepo.getAllFlyerCategories();
+    expect(getFlyersResponse.sort()).toEqual(['Academic', 'Dance']);
+  });
+
+  test('When every flyer has a unique CategorySlug, ensure that all unique CategorySlugs are returned', async () => {
+    const flyerSlugs = ['Academic', 'Dance', 'Social'];
+    const generatedFlyers: Flyer[] = [];
+    for (const slug of flyerSlugs) {
+      generatedFlyers.push(...(await FlyerFactory.createSpecific(1, { categorySlug: slug })));
```
Archit404Error, line 316:
```suggestion
      generatedFlyers.concat(await FlyerFactory.createSpecific(1, { categorySlug: slug }));
```

Archit404Error (dismissed):
Good to go once you clean up these two lines

Archit404Error (approved):
Good to go

vinnie4k (approved):
Good to go.

PR #121, opened 2023-10-21
Change Create Flyer route to use form data for image upload

zachseidner1:
## Overview

Added Express route using form-data to post Flyers. Wasn't sure where to put it so I just put it in `app.ts` so I had easy access to the app. This could definitely be the wrong place though and I want feedback on this. Also I know the code may not be the cleanest, and obviously there's still stuff leftover from the previous create flyer endpoint that is now going unused, so we can either clean everything up later or you all can advise me now in this PR.

## Changes Made

- Added post endpoint for `/flyers/` that takes an image file as form data and puts it on the upload service.

## Test Coverage

I used Postman to test this feature and it worked as expected. I tried removing required fields, tried sending and not sending the file, everything works as expected to my knowledge.

## Next Steps 
- Organize this route in the right place in the code base.
- Clean up any remaining code related to creating flyers (remove it if it's being unused)

image \
Not sure why the image removal is failing but that's not really relevant for what we are trying to do, I think that was failing before I made changes?
image


src/app.ts
```diff
@@ -71,6 +76,64 @@ const main = async () => {
...
+    ) {
+      return res.status(400).json({ error: 'Missing a required field' });
+    }
+
+    // Get the file from form-data and await the upload service
+    const imageB64 = req.file.buffer.toString('base64');
+    const imageURL = await utils.uploadImage(imageB64);
+
+    // Find the organization related to the request
+    const organization = await OrganizationModel.findById(new ObjectID(organizationID));
```
Archit404Error, line 118: You should use the Organization Controller here instead, per MVC design pattern

src/app.ts
```diff
@@ -71,6 +76,64 @@ const main = async () => {
...
+      categorySlug == null ||
+      endDate == null ||
+      flyerURL == null ||
+      location == null ||
+      organizationID == null ||
+      startDate == null ||
+      title == null
+    ) {
+      return res.status(400).json({ error: 'Missing a required field' });
+    }
```
Archit404Error, line 111: You can store the required fields as strings in an array and use `.every()` on the array to check if the value is contained in the object's keys. An example of this is [here](https://stackoverflow.com/questions/55871220/how-to-check-if-a-json-doc-has-all-required-keys-in-node-js).

I honestly think what you have is fine, though, it's probably more readable than the other approach

src/app.ts
```diff
@@ -71,6 +76,64 @@ const main = async () => {
...
+
+    // Get the file from form-data and await the upload service
+    const imageB64 = req.file.buffer.toString('base64');
+    const imageURL = await utils.uploadImage(imageB64);
```
Archit404Error, line 115: Have you checked that the upload service can handle a massive Base64 string sent via a network request (i.e., have you tried uploading a really large file and seeing if that works)

src/app.ts
```diff
@@ -71,6 +76,64 @@ const main = async () => {
     res.json({ success: 'true' });
   });
 
+  const storage = multer.memoryStorage();
+  const upload = multer({ storage: storage });
+
+  // Not really sure where in the code base go, but here's the working route!
+  // I did my part :)
+  // Is just using `/flyers/` the right route here?
```
Archit404Error, line 84: Yeah `/flyers/` is fine, I think this location is fine for now since I don't foresee us adding more flyer functionality. If we do, we'll be able to justify having a separate folder for flyer code and can refactor this then

Archit404Error (approved):
Great work on this Zach! Just a few small comments, otherwise feel free to remove the comments in the code and you should be good to go! Also in the future I'd suggest a more descriptive PR title (you can still change it before you merge) and also if you do have comments about certain chunks of code rather than putting the comments directly in the code you can just make comments on GitHub about the relevant sections after making the PR

src/app.ts
zachseidner1: I checked it just now, and it crashes the backend as utils throws an unchecked exception. Not really sure what to do considering the upload service is sort of a black box since we're using AppDev's upload service and not s3 or something. Do we need to rewrite AppDev's upload service to use form-data too? I feel like there's not much I can change on Volume's backend to fix this other than wrapping in a try catch.

Archit404Error: Yeah I spoke about this with Vin but we would have to add an endpoint to the upload service that can handle a form request, and then you can just forward req.file as part of a form request to the endpoint in the upload service

src/app.ts
zachseidner1: my bad, didn't realize this existed!

src/app.ts
zachseidner1: right, but I can't do that right now since the upload service can't handle a form right now, so is this ok for right now?

Archit404Error: The issue we're trying to solve is that certain orgs can't upload files because the files are too large right? So if we can't deal with large files then we would still have this issue even if we merge this code right? I think if we want to resolve the file size issue once and for all we would have to make that new endpoint in the upload service (which should hopefully be just a couple lines of code because I think `boto3` which is used in the upload service has support for this).

zachseidner1: I agree for sure, I just wasn't sure if I had access to change the upload service. I found the GitHub repository and started working on a fix. I will also update this PR again to use form data in the post request to the upload service once I am done.

vinnie4k (commented):
Great work! We should definitely take a look at the upload service before merging this to ensure that large B64 strings are handled. We should add an endpoint to our upload service that takes in a form request.

vinnie4k (approved):

PR #124, opened 2023-11-07
Fixed issue with edit flyer

vinnie4k:
## Overview

Fixed issues with editing flyer via the form data HTTP request.

## Changes Made

There was an issue where the server would crash if the user did not provide the field `organizationID`. There were also issues where editing the flyer did not even work at all.

I fixed these issues by using the `EditFlyer` function in the `FlyerRepo` which properly handles the logic for editing a flyer.

## Test Coverage

I used Postman to test as well as my local iOS device. I tested every single field with different combinations and it seemed to work just fine. The uploaded images seem to work as well and it even deletes the old image from our storage. I don’t have any screenshots, but I’m confident that this works as expected.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -168,7 +168,7 @@ const main = async () => {
     }
 
     const oldFlyer = await FlyerModel.findById(flyerID);
-    if (oldFlyer == null) {
+    if (!oldFlyer) {
       return res.status(400).json({ error: 'flyerID not associated with any flyers' });
     }
 
--- a/src/app.ts
+++ b/src/app.ts
@@ -6,6 +6,7 @@ import { buildSchema } from 'type-graphql';
 
 import { ApolloServer } from 'apollo-server-express';
 import multer, { diskStorage } from 'multer';
+import path from 'path';
 import { dbConnection } from './db/DBConnection';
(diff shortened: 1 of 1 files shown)


src/app.ts
```diff
@@ -95,7 +96,7 @@ const main = async () => {
       startDate
       title
    * There also must be a file with the key `image` which should be the image
-      associated with the flyer.
+      associated with the flyer. The image type must be png.
```
zachseidner1, line 99: Why do you think this is true? I am pretty sure I updated the upload image function to accept any type of file.

zachseidner1 (approved):
Thanks so much for updating this Vin! I was going to get to it but I have been super busy preparing for prelims and stressed about my workload. I really appreciate you taking on this fix, and sorry for not using the FlyerRepo to edit the Flyer, for some reason I wasn't able to find that method when I was coding the endpoint. Changes look great! Just see my comment about file type.

src/app.ts
vinnie4k: When I used a JPEG, the file extension gets cut off to .jpe. I haven’t looked at the code in the upload service yet to see why this is the issue but perhaps you might know something about it.

PR #126, opened 2023-11-12
Adding routes and migration script for bookmarking

Aayush-Agnihotri:
## Overview
Created new routes for bookmarking articles, flyers, and magazines. Removed the previous "numBookmarkedArticles" field from the user model. Added bookmarkedArticles, bookmarkedMagazines, and bookmarkedFlyers fields to the user model. Created migration script to migrate the database. Removed "numBookmarkedArticles" field from WeeklyDebrief model and refactored associated code to prevent any errors.

## Screenshots
Running "migrate-mongo up" results in
image

Running "migrate-mongo down" results in
image

GraphQL Playground tests
image


vinnie4k (dismissed):
Great work! I'll approve this for now since @cindy-x-liang needs this to work. Can you write a few test cases as well for the routes? I'm not sure if you tested with the GraphQL playground, but I would test with that to make sure the routes are working.

migrations/update-users-model.js
zachseidner1, line 1: I think you should be able to do this using `updateMany` like I provided in the example. This way you don't have to get all the users and do a loop through the collection.

src/repos/UserRepo.ts
```diff
@@ -193,19 +193,116 @@ const incrementShoutouts = async (uuid: string): Promise<User> => {
...
+ */
+const unbookmarkArticle = async (uuid: string, articleID: string): Promise<User> => {
+  const user = await UserModel.findOne({ uuid });
+  if (user) {
+    const articleSlugs = user.bookmarkedArticles.map((article) => article.id);
+    const articleIndex = articleSlugs.indexOf(articleID);
+
+    if (articleIndex === -1) return user;
+
+    user.followedOrganizations.splice(articleIndex, 1);
```
zachseidner1, line 266: Why are you doing this for `followedOrganizations` for bookmarking articles? Shouldn't you use `bookedmarkedArticles`?

src/repos/UserRepo.ts
```diff
@@ -193,19 +193,116 @@ const incrementShoutouts = async (uuid: string): Promise<User> => {
...
+ */
+const unbookmarkMagazine = async (uuid: string, magazineID: string): Promise<User> => {
+  const user = await UserModel.findOne({ uuid });
+  if (user) {
+    const magazineSlugs = user.bookmarkedMagazines.map((magazine) => magazine.id);
+    const magazineIndex = magazineSlugs.indexOf(magazineID);
+
+    if (magazineIndex === -1) return user;
+
+    user.followedOrganizations.splice(magazineIndex, 1);
```
zachseidner1, line 283: Similar to above comment

src/repos/UserRepo.ts
```diff
@@ -193,19 +193,116 @@ const incrementShoutouts = async (uuid: string): Promise<User> => {
...
+ */
+const unbookmarkFlyer = async (uuid: string, flyerID: string): Promise<User> => {
+  const user = await UserModel.findOne({ uuid });
+  if (user) {
+    const flyerSlugs = user.bookmarkedFlyers.map((flyer) => flyer.id);
+    const flyerIndex = flyerSlugs.indexOf(flyerID);
+
+    if (flyerIndex === -1) return user;
+
+    user.followedOrganizations.splice(flyerIndex, 1);
```
zachseidner1, line 300: Similar to above comment

zachseidner1 (changes requested):
Nice work! Looks like you've learned everything pretty quickly :) 
As Vin said, it would be good to add some unit tests for these routes. Also make sure to address my comment on using `followedOrganizations`! I'm worried your code may not work as expected. What testing have you done so far?

vinnie4k (dismissed):
Write some test cases and if it works then you should be good to go. If possible, I want this out ASAP so that Cindy can start working on notifications.

zachseidner1 (dismissed):
Looks good. Please make sure you have tested the code in some way before merging! But I understand we want to get this out quickly so if you merge without writing unit tests it's ok. Just make sure you have tested the code _somehow_

vinnie4k (approved):
Good to go.

